PR #176, opened 2026-05-02
test: pass-7 coverage — progress + term + repl (+70 tests, 2 fixes)

ser-vasilich:
## Summary

Three C-side test suites driving REPL-adjacent files that the rfl harness can't reach (the harness calls \`ray_eval_str\` directly, bypassing both the line-edit terminal and the repl wrapper).  Plus two latent bug fixes that the new tests surfaced.

### Bug fixes

1. **\`ray_term_visual_width\` ANSI escape state machine** (\`src/app/term.c\`).  Single-state machine treated the CSI introducer \`[\` (0x5B) as the final byte because 0x5B sits in the [0x40, 0x7E] "final byte" range.  Result: every CSI sequence (SGR colors, cursor moves) terminated one byte early, and the digits/\`;\`/\`m\` characters that followed leaked into the visual-width count.  For \`"\\033[31mred\\033[0m!"\` the function returned 9 instead of 4.  The REPL under-counted rows on multi-line input and mis-positioned the cursor whenever syntax highlighting was active.  Fix: 3-state machine.

2. **Comments-only file in \`ray_repl_run_file\`** (\`src/app/repl.c\`).  A file containing only \`;;\` line comments would cause \`parse_source\` to error with "parse" because it expected at least one expression — but every Lisp/shell-style language treats a header-only file as a successful no-op.  Fix: single-pass whitespace+comment scan that short-circuits to \`return 0\`.

### Coverage delta

| File | Before | After | Δ |
|------|-------:|------:|--:|
| \`src/core/progress.c\` | 10.84% | **100.00%** | +89.16pp |
| \`src/app/term.c\` | 0.00% | **46.28%** | +46.28pp |
| \`src/app/repl.c\` | 0.00% | **26.89%** | +26.89pp |
| **TOTAL line coverage** | 75.18% | **76.96%** | +1.78pp |

\`term.c\` and \`repl.c\` were both at 0% — the rfl harness goes through \`ray_eval_str\`, never the line-edit code or the repl wrapper, so without dedicated C tests these files had no direct coverage.  46% / 27% reflects the testable subsets (history, cursor math, ANSI escape printers; \`ray_repl_run_file\` + piped-mode loop); the remaining ~50% / ~70% is interactive raw-mode keystroke handling that needs a real PTY to exercise — out of scope here.

### Suites added

- \`test_progress.c\` (15 tests): callback fires after \`min_ms\`, \`tick_ms\` throttle, label/end semantics, snapshot fields, lazy start, user-pointer plumbing, end-without-callback.
- \`test_term.c\` (34 tests): cursor + line escape printers, \`ray_term_get_size\` ioctl + 80x24 fallback, \`ray_term_visual_width\` (ASCII / ANSI / 1..4-byte UTF-8 / SS3 / bare ESC), paren-match, history (create/add/nav/search/save+load round-trip), bracket-balance state across chunks, \`ray_term_redraw\` (bracket-match + multiline cont prompt), \`ray_term_collect_completions\` (env + \`from:\` table cols + history words).
- \`test_repl.c\` (21 tests): \`ray_repl_run_file\` happy / multi-form / parse-error / eval-error / empty / comments-only / nonexistent / multiline / lazy result; lifecycle (\`create\` null-poll, \`destroy\` null); piped-mode loop via stdin pipe redirection (single, multi-line, multi-form, empty, comments-blank, \`\\\` exit, \`exit\` word, \`:quit\`, command, error-continue).

## Test plan

- [x] \`make test\` — 1224 of 1225 passed (1 pre-existing skip, 0 failed)
- [x] ASan/UBSan clean
- [x] Coverage measured both with and without pass-7 to compute the delta

### DIFF
--- a/src/app/repl.c
+++ b/src/app/repl.c
@@ -1186,6 +1186,26 @@ int ray_repl_run_file(const char* path) {
 
     if (nread == 0) { ray_release(block); return 0; }
 
+    /* Skip files whose entire content is whitespace and `;` line
+     * comments — `parse_source` raises a `parse` error for empty
+     * input, which is the wrong UX when running `rayforce file.rfl`
+     * on a comments-only file (a script header with no body should
+     * be a successful no-op).  Cheap single-pass scan; mirrors the
+     * lexer's whitespace/comment rule. */
+    {
+        const char* p = buf;
+        const char* end = buf + nread;
+        bool any_expr = false;
+        while (p < end) {
+            char c = *p;
+            if (c == ' ' || c == '\t' || c == '\n' || c == '\r') { p++; continue; }
+            if (c == ';') { while (p < end && *p != '\n') p++; continue; }
+            any_expr = true;
+            break;
+        }
+        if (!any_expr) { ray_release(block); return 0; }
+    }
+
     /* File mode matches eval_and_print's structure so -t 1 / :t 1
      * produces the same span layout (parse / eval / materialize) as
      * REPL input, not one opaque "eval" tick.  Colors are keyed per
--- a/src/app/term.c
+++ b/src/app/term.c
@@ -205,12 +205,28 @@ void ray_term_get_size(ray_term_t* term) {
 
 int32_t ray_term_visual_width(const char* str, int32_t len) {
     int32_t width = 0;
+    /* 0=normal, 1=just saw ESC (next byte is the CSI/SS3 introducer:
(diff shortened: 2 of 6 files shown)


ser-vasilich:
Closing — bundling pass-7 with deeper repl.c push (target 75%+) and dead-code review into a single subsequent PR.

PR #224, opened 2026-06-03
diagnose: per-test SIGALRM watchdog to identify macOS-debug hang

ser-vasilich:
## What

Adds a SIGALRM-based per-test timeout watchdog to `test/main.c`. When a test
exceeds the timeout (default **90 s**), the SIGALRM handler writes the offending
test name to stderr via async-signal-safe `write(2)` and calls `_exit(124)` —
so CI logs capture the culprit and the job finishes instead of hanging.

## Why a draft PR

PR #223's macOS-debug job hung for 27+ minutes (vs ~150 s on master), suggesting
a platform-specific infinite loop or UB triggered only under macOS + ASan/UBSan.
Without per-test timeout, the suite stalls and we never learn which test is the
culprit. This draft PR isolates the watchdog so CI can identify the hanging test
on its own, then we can investigate the actual bug separately.

The watchdog is opt-out via `RAY_TEST_TIMEOUT_S=0`; the default 90 s comfortably
exceeds the slowest legitimate tests (1.05M-row HLL, splayed I/O round-trips).

## Expected outcome

- ubuntu-{debug,release}, macos-release: green in ~2-3 min as on master
- **macos-debug: should fail with `TIMEOUT in test: <name>` after ≤90 s per test,
  pinpointing which test introduces the hang**

Once we know the test name, we can reproduce locally / inspect for UB and open a
proper fix PR.

### DIFF
--- a/test/main.c
+++ b/test/main.c
@@ -43,6 +43,7 @@
 #include <dirent.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#include <signal.h>
 
 #include <rayforce.h>
 #include "lang/eval.h"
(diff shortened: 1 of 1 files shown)


ser-vasilich:
Closing as a one-off diagnostic — the macOS-debug hang on PR #223 turned out to be a transient CI runner flake (master is green for the same code). The watchdog itself is small and uncontroversial; can be reopened if a future hang makes it worth carrying as standing infrastructure.

PR #265, opened 2026-06-19
ci: fix release workflow tag bootstrap (v2.1.0 to follow via dispatch)

singaraiona:
## What

Fixes the release workflow that failed on the first v2.1.0 merge.

**Root cause:** a *draft* GitHub release does not create its git tag — the tag
ref only appears when the release is published. The `build` job checked out
`ref: vX.Y.Z` *between* draft-create and publish, so `git fetch` found no such
tag and failed; `publish` (which creates the tag) never ran.

## Changes (`.github/workflows/release.yml`)

- `build` now checks out the release's **target commit** (`prepare` output `sha`),
  never the not-yet-existing tag.
- `prepare` is **idempotent**: it reuses an existing draft and builds from *its*
  target, so artifacts always match the tag `publish` creates.
- Adds a **`workflow_dispatch(version)`** trigger as the re-run / first-release
  bootstrap path.
- Bumps `actions/checkout` v4 → v5 (silences the Node 20 deprecation).

## Note on this PR's title

Intentionally **not** a `vX.Y.Z` title, so the (still-broken) release workflow on
master is skipped on merge — no junk release run. The real **v2.1.0** is then cut
by dispatching the fixed workflow: `gh workflow run release.yml -f version=2.1.0`,
which reuses the existing draft and tags v2.1.0 at the original merge commit.

`version-guard` will show red on this PR (title isn't a version) — expected and
non-blocking.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,59 +1,88 @@
 name: Release
 
-# Fires when a release PR (dev -> master, titled vX.Y.Z) is merged. The git
-# tag is the single source of truth: this workflow CREATES the tag from the PR
-# title, builds the optimized binary per platform with the version injected at
-# compile time (make ... RAY_VERSION=X.Y.Z), and publishes a GitHub Release.
-# No source file is bumped and nothing is pushed to the protected master branch
-# — only a tag ref and a release are created, which GITHUB_TOKEN can do without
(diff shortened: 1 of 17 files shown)


singaraiona:
Superseded by #266 — an isolated topic branch carrying only the release-workflow fix. dev has since picked up unrelated feature work (f64 null model, 3302f318) that must not ride to master under this CI fix.

PR #281, opened 2026-06-23
v2.1.5

### LINKED ISSUE #280: .db.splayed.set wrongly a sym column
`.db.splayed.set` wrongly rejects a `sym` column

```
‣ (.db.splayed.set "/tmp/t/x" (table [sym v] (list [1 2] [3 4])))
error: reserve
‣ (.db.splayed.set "/tmp/t/x" (table [Sym v] (list [1 2] [3 4])))
┌─────┬───────────────────────────────┐
│ Sym │               v               │
│ I64 │              I64              │
├─────┼───────────────────────────────┤
│ 1   │ 3                             │
│ 2   │ 4                             │
├─────┴───────────────────────────────┤
│ 2 rows (2 shown) 2 columns (2 shown)│
```

singaraiona:
**Free `sym` as a column name; the splayed symfile becomes the `.sym` dotfile.**

Fixes #280.

## Problem
`.db.splayed.set` rejected any column named `sym` with `error: reserve`:

```
‣ (.db.splayed.set "/tmp/t/x" (table [sym v] (list [1 2] [3 4])))
error: reserve
‣ (.db.splayed.set "/tmp/t/x" (table [Sym v] (list [1 2] [3 4])))   ;; worked
```

The on-disk symfile (`<dir>/sym`) and its lock (`<dir>/sym.lk`) occupied
column-legal names, so a column named `sym` would clobber them — and on read,
`<dir>/sym` is unconditionally treated as the symfile. The name was reserved at
the storage-format level, which made the canonical q `sym` ticker column
unsavable.

## Fix
Rename the symfile to the **dotfile `.sym`** (lock `.sym.lk`) everywhere it is
written/resolved — standalone splayed dir, parted root, CSV splayed/parted
writers, and the parted reader. Dot-led names are already excluded from being
columns (like `.d`), so:

- a plain `sym` column round-trips (the issue's case), and
- a real SYM-typed `sym` column coexists with the `.sym` domain and its
  `.sym.lk` lock.

The reserved-name guard is now **path-based, not name-based**: it fires only
when a symfile is actually written (table has SYM columns) **and** the resolved
`sym_path` would land on a real column file (or its `.lk` lock) in the same dir.
This still catches collisions from an explicit 3-arg `sym_path` while never
false-rejecting the default `.sym`.

Greenfield hard cutover: existing on-disk symfiles named `sym` are not migrated.

## Tests
- Rewrote splay test 33b to cover: (a) the issue's I64 `sym` column, (b) a
  SYM-typed `sym` column round-trip alongside `.sym`/`.sym.lk`, (c) explicit
  `sym_path` colliding with a column → `reserve`, (d) explicit lock collision →
  `reserve`, (e) no false positive for a different-dir symfile.
- Updated rfl integration tests, stress fixtures, and storage docs to the
  `.sym` convention.
- Full suite: **3487 passed, 2 skipped, 0 failed** (ASan/UBSan).


singaraiona:
Landed on `dev` directly (47774192) per request — not merging to master at this time. Closing this master-targeting PR; the fix for #280 is on dev.

PR #307, opened 2026-07-06
feat(store): embeddable append-only log (AOF) C API

vsov:
## Why

Host applications embedding `librayforce` (concretely: the starwatcher trading platform, which wants Rayforce as its event-log + analytics layer) need a **durable append log**: opaque payloads, monotonic contiguous LSNs, an explicit durability barrier, segment rotation, and crash recovery.

Neither existing facility fits:

- **Transaction journal** (`store/journal.h`) — wired into the IPC dispatch loop only, entries are replayable eval frames, torn tail is deliberately fatal (entries are commands with side effects), single-threaded by design.
- **Column save** (`store/col.h`) — whole-column `tmp + fsync + rename`; durable but not incremental.

## What

New `src/store/aof.{c,h}` + public API section in `include/rayforce.h`:

```c
ray_aof_t* ray_aof_open(const char* dir, int64_t segment_limit, ray_err_t* out_err);
int64_t    ray_aof_append(ray_aof_t*, const void* payload, int64_t len, ray_err_t* out_err); // buffered, returns LSN
ray_err_t  ray_aof_commit(ray_aof_t*);                    // fsync barrier
int64_t    ray_aof_next_lsn(const ray_aof_t*);
int64_t    ray_aof_scan(const char* dir, int64_t from_lsn, ray_aof_scan_cb_t cb, void* ctx, ray_err_t* out_err);
ray_err_t  ray_aof_close(ray_aof_t*);
```

- On-disk: one dir per log, segments named `%020lld.aof` by first LSN, record = `u32 LE len + u32 LE CRC32 + payload`, rotation past the configured byte limit (default 256 MiB), new-segment dir entry fsync'd (same discipline as `col.c`).
- **Torn tail ≠ corruption**: a torn tail is data the writer was never acknowledged for — truncated on open, LSN sequence continues. This deliberately differs from the journal and is documented in both headers. A bad record in the **interior** of the log *is* corruption: scans fail with `RAY_ERR_CORRUPT`.
- Threading: one writer handle per dir (no internal locking); concurrent scans are safe against a live writer — they observe the committed prefix and stop cleanly at the tail.
- Independent of `ray_runtime_t` lifecycle (usable before create / after destroy); uses `ray_sys_alloc` + `store/fileio.h` helpers; follows `journal.c`'s stdio + `fsync(fileno())` pattern.

## Tests

`test/test_aof.c`, 9 cases registered in `test/main.c`, **no runtime fixture** (doubles as proof of runtime-independence): roundtrip, reopen-continues-LSN, torn-tail truncation, uncommitted-invisible-to-scan, rotation (+ reopen mid-history), scan-from with segment skipping, early-stop callback, interior-corruption detection, empty-payload/bad-args.

Full suite: **3560 of 3561 passed (1 skipped — pre-existing, 0 failed)**; `make lib` links `aof.o` into `librayforce.a` cleanly.

## Out of scope / follow-ups

- Windows parity (POSIX `opendir`/`truncate`/`fileno` — same scope as `journal.c` today).
- Rayfall ops surface (`.aof.open` / `.aof.write` verbs) — C API first; verbs can wrap it if wanted.
- Retention helpers (delete-segments-before-LSN) — trivial to add when a consumer needs them.

### DIFF
--- a/include/rayforce.h
+++ b/include/rayforce.h
@@ -691,6 +691,69 @@ ray_t*    ray_ipc_send(int64_t handle, ray_t* msg);
 ray_err_t ray_ipc_send_async(int64_t handle, ray_t* msg);
 ray_t*    ray_ipc_send_verbose(int64_t handle, ray_t* msg);
 
+/* ===== Append-Only Log (AOF) API =====
+ *
+ * Embeddable durable event log for host applications: opaque payloads,
+ * monotonic contiguous LSNs, explicit commit barriers, segment rotation,
+ * crash recovery.  Independent of the ray_runtime_t lifecycle — usable
+ * before ray_runtime_create and after ray_runtime_destroy.
+ *
+ * Durability model: ray_aof_append() buffers and returns the reserved
+ * LSN; ray_aof_commit() is the fsync barrier — every append issued
+ * before it is durable once it returns RAY_OK.  A crash discards only
+ * un-committed appends: on the next open the torn tail is truncated and
+ * the LSN sequence continues from the last committed record.  (This is
+ * deliberate and differs from the transaction journal, where a torn
+ * tail is fatal: AOF payloads are opaque data the writer was never
+ * acknowledged for, not replayable commands.)
+ *
+ * Threading: one writer handle per directory, no internal locking.
+ * Concurrent ray_aof_scan() calls are safe, including against a live
+ * writer — a scan observes the committed prefix and stops cleanly at
+ * the first incomplete tail record.
+ *
+ * Errors: RAY_ERR_DOMAIN on bad arguments, RAY_ERR_IO on filesystem
+ * failure, RAY_ERR_LIMIT on payload > UINT32_MAX, RAY_ERR_CORRUPT when
+ * a scan finds a bad record in the interior of the log (not the tail).
+ */
+
+typedef struct ray_aof_s ray_aof_t;
+
+#define RAY_AOF_DEFAULT_SEGMENT_LIMIT ((int64_t)256 * 1024 * 1024)
+
+/* Open (creating the directory if needed) and recover the log at `dir`.
+ * `segment_limit` <= 0 selects RAY_AOF_DEFAULT_SEGMENT_LIMIT.  On
+ * failure returns NULL and sets *out_err (out_err may be NULL). */
+ray_aof_t* ray_aof_open(const char* dir, int64_t segment_limit,
+                        ray_err_t* out_err);
+
+/* Buffered append of `len` bytes; returns the reserved LSN (>= 0), or
+ * -1 with *out_err set.  Durable only after ray_aof_commit(). */
+int64_t    ray_aof_append(ray_aof_t* log, const void* payload,
+                          int64_t len, ray_err_t* out_err);
+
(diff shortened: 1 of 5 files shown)


vsov:
Audit findings addressed in d0c5a831.

**Blocker 1 (committed-prefix contract / LSN reuse)** — fixed by moving the commit boundary INTO the stream: `ray_aof_commit()` now writes a sentinel commit frame (len `0xFFFFFFFF`, CRC field = fixed magic `0xB0AD63C9` = CRC32 of `"ray-aof-commit-frame"`) before flush+fsync. Scans deliver only records below the last valid frame, so uncommitted appends are invisible **by construction** — regardless of stdio buffering or page cache — and recovery truncates strictly after the last frame, so a reissued LSN can only ever belong to a record no scan could have observed. Damage *below* the boundary now fails `ray_aof_open`/`ray_aof_scan` with `RAY_ERR_CORRUPT` instead of silent truncation (acknowledged data is never dropped quietly). Still one fsync per commit; rotation remains a commit point. The misleading test was replaced: `aof/uncommitted_invisible` now appends 512 KB (past any stdio buffer) and two new tests cover the audit's exact scenarios: `aof/crash_lsn_reuse_unobservable`, `aof/committed_corruption_open`.

**Blocker 2 (CRC init race)** — the table is now a compile-time `static const` array; the racy lazy init is gone entirely.

**Non-blocking, also fixed:** record CRC now covers length+payload (a bit-flipped length can't fake a ~4 GB record during recovery); framing is explicit little-endian on all architectures (matching the docs); segment names must be exactly 20 digits + `.aof`; payload cap moved one byte below the frame sentinel (`RAY_ERR_LIMIT` at ≥ 2³²−1). Tail-recovery semantics (bit-rot vs torn tail) are now documented in both headers — and with commit frames the distinction is real: CRC failure below the boundary = corrupt, bytes above it = torn.

**On the failing `build-and-test` check:** `exec/streaming_large_dag` crashes with an ASan stack-overflow in `exec_node_inner` (exec.c:1601) on a **clean `origin/dev` worktree, 3/3 runs, zero AOF code present**, and today's dev "Nightly stability" run is also red — pre-existing dev breakage, unrelated to this PR. All 11 `aof/*` tests pass. Filed separately for a fix.

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

vsov:
**On the `static-analysis` check:** the runner's cppcheck was updated and now flags ~10 files across dev (`term.c`, `compile.c`, `format.c`, `exec.c`, `expr.c`, `fused_topk.c`, `group.c`, `idxop.c`, `pivot.c`, `fileio.c`) — **zero findings in the AOF diff** (`grep -c aof` on the job log = 0). The `fileio.c:260` FP adjacent to this PR's usage was suppressed in 26102444 per repo convention; the remaining findings are dev-wide, pre-date this PR, and belong in a dev-branch cleanup rather than here. Note the Makefile's own stance: "cppcheck is a second-opinion linter — advisory only, never a gate" — and `ci-success` (the merge gate) is green, as are the audit, all four builds, tsan and fuzz-smoke.

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

PR #334, opened 2026-07-18
fix(store): read full link sidecar to avoid wrong-symbol truncation

belowzeroff:
## What & why

`try_load_link_sidecar` (`src/store/col.c`) read a linked column's target table
sym name into a fixed 256-byte stack buffer (`fread` of 255 bytes). A target
name longer than 255 bytes was silently truncated, so `ray_sym_intern` interned
a **different** symbol and the loaded column linked to the **wrong table** —
silent data corruption across a save/load round-trip. The writer side already
emits the full, untruncated name, so only the reader was capping.

(The `255` limit in `sym.c` bounds the number of dot-separated *segments*, not
the byte length of a name, so names longer than 255 bytes are legal and
reachable.)

The fix reads the whole sidecar into a buffer sized to the file, capped at 1 MiB
so a corrupt/oversized sidecar can't force a large allocation, using the
`ray_alloc_raw` / `ray_free_raw` pair already used elsewhere in the file. The
trailing-whitespace trim and best-effort semantics (any failure leaves the
column unlinked) are preserved.

### Test

`link/persistence_long_target_name` links a column through a 300-byte target
name, saves, loads, and asserts the loaded `link_target` still resolves to the
original symbol. Verified as a genuine regression test — with the old reader the
suite reports:

```
link/persistence_long_target_name   FAIL
=== 3631 of 3633 passed (1 skipped, 1 failed) ===
```

and with the fix it passes.

Full suite under the debug ASan + UBSan build: `3632 of 3633 passed (1 skipped,
0 failed)`.

## Checklist

- [x] Commits follow Conventional Commits (`fix:`)

### DIFF
--- a/src/store/col.c
+++ b/src/store/col.c
@@ -580,14 +580,28 @@ static void try_load_link_sidecar(ray_t* vec, const char* path) {
     memcpy(link_path + plen, ".link", 6);
     FILE* lf = fopen(link_path, "rb");
     if (!lf) return;
-    char buf[256];
-    size_t n = fread(buf, 1, sizeof(buf) - 1, lf);
+    /* Read the whole sidecar.  The target sym name can be longer than any fixed
+     * buffer, and a truncated read would intern a DIFFERENT symbol — silently
+     * linking the column to the wrong table.  Size to the file, capped so a
+     * corrupt/oversized sidecar can't force a huge allocation. */
+    if (fseek(lf, 0, SEEK_END) != 0) { fclose(lf); return; }
+    long fsz = ftell(lf);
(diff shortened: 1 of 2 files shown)


singaraiona:
Please reject a short read before trimming or interning. `fread` may return `0 < n < fsz` on an I/O error or if the sidecar is truncated after `ftell`; the current code then interns that prefix and sets `HAS_LINK`, recreating the wrong-symbol linkage this patch is intended to prevent. After closing the file, check `n == (size_t)fsz`; otherwise free `buf` and return.

belowzeroff:
Thank you. Added an explicit n == fsz short-read check before trimming/interning; a partial read now frees the buffer and leaves the column unlinked. Amended in 1c63994a.

PR #338, opened 2026-07-23
perf: parallelize serial stages around group-by (binary aggs, OP_IF, filter/where/gather)

ser-vasilich:
# perf: parallelize the serial stages around group-by (binary aggs, OP_IF, filter/where/gather)

## Problem

On a 48-core machine the TAQ benchmark (medium NYSE dataset, quote = 296.6M rows) showed group-by queries scaling only **2.1–2.3×** from 1 to 48 cores. Profiling the full query pipeline showed the group-by kernel itself was fine after per-worker accumulation — the wall-clock was dominated by *serial* stages around it:

- binary aggregates (`pearson`/`cov`/`wavg`/`wsum`) were not DA-eligible → fell to slower paths;
- `OP_IF` projections went through `exec_if_selected` (serial per-row id-scaffolding) even when no WHERE was present;
- bitmap→index materialization in `exec_filter` / `sel_compact` was serial;
- the `where` builtin and `at col idx` gather were serial;
- `ray_pool_dispatch` was being fed chunk counts, so 1500+ chunks morselized into ~2 tasks.

## What changed

**`src/ops/group.c`** — binary aggregates via the DA (dense-array) path: per-worker accumulators extended with co-moments (Σx, Σx², Σy, Σy², Σxy), finals computed in `emit_agg_columns`. Two merge bugs fixed along the way: Σx for integer x-columns is stored as double but was merged through the `int64` union member (garbage like 1.3e305 at high core counts); the `da_merge_fn` path didn't merge the PAIR co-moment arrays at all (wrong pearson with ≥1024 slots). Eligibility is gated: y-column must be FP or non-nullable integer — nullable-int y falls back, since DA reads raw values and would sum null encodings as data.

**`src/ops/pivot.c`** — `OP_IF`: (1) the eager elementwise fill is parallel (type-switched morsel fill; SYM LUT warm-up stays serial before dispatch; STR output stays serial); (2) trivial-branch ifs (scan/const through alias/materialize) are rerouted from `exec_if_selected` to the eager path, gated by an out-type whitelist — `exec_if_selected` fires even without a WHERE clause and was the hidden per-row serial cost in the group-by queries' projections.

**`src/ops/filter.c`** — `exec_filter` and `sel_compact`: bitmap→`match_idx` build parallelized with 3-phase compaction (per-chunk popcount → serial exclusive prefix sum → per-chunk fill at disjoint offsets), one task per chunk via `ray_pool_dispatch_n`. Chunk count capped at 1024 (the initial task-ring capacity), so ring growth can never fail and silently drop tasks.

**`src/ops/builtins.c` / `src/lang/eval.c`** — `where` builtin parallel with the same 3-phase pattern (keeps the word-at-a-time zero-skip in the emit phase); `gather_by_idx` parallel with a `ray_parallel_flag` reentrancy guard (existing callers may already be inside a dispatch); null-bit propagation stays serial.

All parallel paths share the same gates: `pool && pool->n_workers > 0` (at `-c 1` the pool exists with 0 workers), `len >= RAY_PARALLEL_THRESHOLD`, serial fallback on any allocation failure.

## Results

48-core box, canonical `queryEngines.sh`, medium TAQ, min of 3 runs, **both binaries built from the same dev base** (`8f8033c`):

| Query | dev c48 (ms) | branch c48 (ms) | speedup | scaling 1→48 before | after |
|---|--:|--:|--:|--:|--:|
| Q22 (wavg by ex) | 3932 | **1443** | 2.72× | 2.1× | **5.6×** |
| Q23 | 3731 | **1405** | 2.66× | 2.2× | **5.7×** |
| Q24 (by sym,ex) | 6180 | **2293** | 2.70× | 2.3× | **5.9×** |
| Q25 | 5479 | **1770** | 3.10× | 2.2× | **6.7×** |

Side wins from the filter/where/if work: Q48 2.28×, Q42 1.69×, Q47 1.63×, Q7 1.49×, Q39 1.29×, Q8/Q16/Q17/Q19/Q41 1.2–1.3×.

**Single-core: no regressions** — branch ≤ dev at c1 across all 53 queries (e.g. Q22: 8052 vs 8216 ms), thanks to the `n_workers > 0` gates.

## Correctness

- `make test`: **3639/3639** on a clean build.
- `wavg`/`pearson` results verified against numpy to full precision at c1/c4/c24, including FP-y columns with NaNs and integer x-columns.
- Edge cases that can't be handled safely (nullable-int y, STR outputs, mixed-type if branches, reentrant dispatch) fall back to the existing serial paths via explicit gates.

## Known remaining gaps (out of scope here)

- Q22-25 still have headroom at high core counts: each op (`in`, `if`, gather, arithmetic) materializes its full result before the next starts. Fusing the filter/projection expressions into one compaction pass would remove most of the remaining wall-clock — architectural follow-up.
- Window functions parallelize across partitions only (`window.c` Pass 3); low-cardinality partitioning (e.g. by exchange = 19 parts) caps their scaling.

https://claude.ai/code/session_01JZwH1gpAeLzDX2bxc4jz6s


ser-vasilich:
Audit findings addressed — point by point:

**Blocker 1 (y-side null semantics diverge between paths).** Fixed by unifying, and the fix went further than the suggested minimum. The legacy HT accumulators now implement the same pair-skip contract as the scalar reducer, the v2 engine and the DA path: a null on either side voids the whole (x,y) pair (`accum_from_entry_nullable` skips before `nn++`/Σx; entry packing canonicalizes integer nulls — NaN in F64-packed slots, `NULL_I64` in int×int slots — so the accumulator needs no per-agg y sentinel plumbing). One precision on the finding's premise: a plain-scan SYM key does not reach the legacy HT — `agg_v2_can_handle` admits SYM scans, and v2 already pair-skips. The legacy HT is reached via expression keys, which is where the divergence was real and is now closed.

Chasing this exposed two more latent bugs in the same code, both fixed:
- both HT emitters (radix and serial) divided pearson/cov/scov by the group row count (`cnt`) instead of the accumulated pair count (`nn`) — wrong moments whenever a group contains any null on either side, independent of this PR's changes;
- the all-null-group guards wrote `v = 0.0` before the common store, which overwrote the null sentinel that `ray_vec_set_null` had just written — an all-null group emitted `0.0` instead of a typed null.

**Blocker 2 (no tests).** Added `test/rfl/agg/binary_group_null_paths.rfl`: 46 assertions covering wsum/wavg/pearson_corr/cov/scov on all three grouped routes — v2 (plain-scan int key), DA (expression int key, y-eligible), legacy HT (expression key + nullable-int y, both the F64-packed and int-packed entry lanes) — against independently computed pair-skip truth, for all four x/y type combinations (F64×F64, I64×I64, I64×F64, F64×I64) with nulls on both sides, plus an all-pairs-null group asserting `wsum = 0.0` and typed nulls for the ratio/moment aggs on every route.

**3 (inconsistent nested-dispatch guard).** Added `ray_par_dispatch_ok()` to `ops/internal.h` (workers + threshold + `ray_parallel_flag` reentrancy check) and applied it at all four sites; `builtins.c` carries a local copy (it cannot include `ops/internal.h` — `clear_neg_zero` clash — noted in a comment).

**5 (dead `agg_pair_mask`).** Removed.

**6 (wasted parallel count for vector inputs).** The premise is not quite right: `pass_count` from the parallel count phase is consumed by `exec_filter_vec` (it sizes the output and saves the sequential counting sweep). The only discarded work is the chunk-offset prefix — ≤1024 additions. Left in place; added a comment saying so.

**7 (threshold inconsistency).** Both the gate and the table fallback now derive from one value (`fidx_rows`).

**8 (short y OOB reachable via DA).** The DA eligibility gate now rejects `yv2->len < n_scan`.

`make test`: 3640/3640 (the new test included) on a clean ASan/UBSan build.

ser-vasilich:
Superseded by #341 — the audit-gate blockers (y-side null semantics + missing regression tests) are fixed there, and the description was reworked.

PR #339, opened 2026-07-23
fix(expr): avoid null truthiness casts in fallback binary ops

belowzeroff:
## Problem

`binary_range`'s fallback (non-fused) kernels for `OP_AND` / `OP_OR` cast the widened `double` operand straight to `uint8_t`:

```c
uint8_t li = (uint8_t)LV_READ(i);
```

`LV_READ` widens integer operands to `double` and yields `NaN` for float nulls. Casting a `NaN` — or an out-of-range value such as a widened `NULL_I64` (`-9.2e18`) — to `uint8_t` is undefined behavior (C11 §6.3.1.4).

UBSan, from `make test` (`-fsanitize=address,undefined`):

```
src/ops/expr.c: runtime error: -9.22337e+18 is outside the range of representable values of type 'unsigned char'
src/ops/expr.c: runtime error: nan is outside the range of representable values of type 'unsigned char'
```

Reproduced by `expr_null/diff_i64_and_raw`, `expr_null/diff_i64_or_raw`, `expr_null/diff_f64_andor_chokes`.

## Fix

Introduce two branch-free truthiness helpers and route AND/OR through them:

- `truthy_i64ish(v)` — treats `0` and the widened i64 null sentinel as false.
- `truthy_f64ish(v)` — treats `0.0` and `NaN` (float null) as false.

No value is ever cast out of range, so the UB is gone. Non-null truthiness is unchanged, and the existing `fix_null_comparisons` post-pass still forces null-input positions to `0` — so results stay bit-identical to the fused kernel (asserted by the `diff_*` differential tests, which compare the fused and fallback paths).

## Verification

- `make test` under `-fsanitize=address,undefined`: the three target tests pass with **zero** UBSan runtime errors.
- Full suite: no regressions.

### DIFF
--- a/src/ops/expr.c
+++ b/src/ops/expr.c
@@ -29,6 +29,20 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+/* Truthiness for the integer AND/OR fallback lane.  `LV_READ`/`RV_READ` widen
+ * the operand to double, so a null arrives as the per-type sentinel cast to
+ * double — (double)NULL_I16, (double)NULL_I32, or (double)NULL_I64 depending on
+ * the bound operand pointer.  `nullv` is that sentinel; a null or zero operand
+ * is false.  Comparing the widened double (never casting it back to an integer)
+ * also sidesteps the UB of (uint8_t)NaN and (uint8_t)(out-of-range double). */
+static inline uint8_t truthy_intish(double v, double nullv) {
+    return (v != 0.0 && v != nullv) ? 1 : 0;
+}
+
+static inline uint8_t truthy_f64ish(double v) {
+    return (v == v && v != 0.0) ? 1 : 0;
+}
+
 static bool atom_to_numeric(ray_t* atom, double* out_f, int64_t* out_i, bool* out_is_f64) {
     if (!atom || !ray_is_atom(atom)) return false;
     switch (atom->type) {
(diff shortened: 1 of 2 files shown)


src/ops/expr.c
```diff
@@ -29,6 +29,20 @@
...
 #include <stdlib.h>
 
+/* Truthiness for the integer AND/OR fallback lane.  `LV_READ`/`RV_READ` widen
+ * the operand to double, so a null arrives as the per-type sentinel cast to
+ * double — (double)NULL_I16, (double)NULL_I32, or (double)NULL_I64 depending on
+ * the bound operand pointer.  `nullv` is that sentinel; a null or zero operand
+ * is false.  Comparing the widened double (never casting it back to an integer)
+ * also sidesteps the UB of (uint8_t)NaN and (uint8_t)(out-of-range double). */
+static inline uint8_t truthy_intish(double v, double nullv) {
+    return (v != 0.0 && v != nullv) ? 1 : 0;
```
singaraiona: `LV_READ` has already widened the original `int64_t` to `double` here, so comparing it with `(double)NULL_I64` cannot distinguish every valid I64 value from the sentinel. For example, both `INT64_MIN + 1` and `INT64_MIN + 2` round to the same binary64 value as `INT64_MIN`; `truthy_intish()` therefore marks these valid non-zero values false. The later `fix_null_comparisons()` pass does not correct them because the source rows are not null. I reproduced this by adding those values to the raw I64 fixture: `diff_i64_and_raw` reported fused=1/fallback=0, and `diff_i64_or_raw` did the same when paired with zero. Please perform truthiness/null detection without collapsing the original I64 payload through `double` (or rely on the existing null post-pass while testing only zero here).

singaraiona (commented):
Found a correctness issue in the new integer truthiness helper: valid near-sentinel I64 values make the fallback path disagree with the fused path.

PR #340, opened 2026-07-23
fix(null): avoid f64 null casts to integers

belowzeroff:
## Problem

Several paths cast an F64 value straight to an integer type. When the F64 holds a null — encoded as `NaN` (`NULL_F64`) — the cast is undefined behavior (C11 §6.3.1.4: `NaN` / out-of-range float → integer is UB).

UBSan, from `make test` (`-fsanitize=address,undefined`):

```
src/ops/builtins.c: runtime error: nan is outside the range of representable values of type 'long long'
src/ops/expr.c:     runtime error: nan is outside the range of representable values of type 'int'
src/ops/query.c:    runtime error: nan is outside the range of representable values of type 'long long'
```

Reproduced by `expr_null/diff_cast_f64_to_i32`, `rfl/null/cast`, `rfl/expr/const_expr`, `compile/update_promo_f64_to_i64_null_slot_is_sentinel`.

## Fix

Add null-aware cast helpers in `core/types.h` and use them at the affected sites (`builtins.c` cast worker, `expr.c` const-fold + unary/elementwise casts, `query.c` update promotion):

- `ray_cast_f64_to_i64_null` — `NaN` → `NULL_I64`, plus saturation so out-of-range finite values never land on the sentinel.
- `ray_cast_f64_to_i32_null` / `_i16_null` — `NaN` → `NULL_I32` / `NULL_I16`.
- `ray_cast_f64_to_u8_null` / `_bool_null` — `NaN` → `0` (U8 / BOOL are non-nullable).

Final values are unchanged (null positions were already sentinel-masked); this removes the UB and writes the canonical null sentinel in-buffer.

## Verification

- `make test` under `-fsanitize=address,undefined`: the four target tests pass with **zero** UBSan runtime errors.
- Full suite: no regressions.

### DIFF
--- a/src/core/types.h
+++ b/src/core/types.h
@@ -50,15 +50,23 @@ static inline int64_t ray_cast_f64_to_i64_null(double v) {
 }
 
 static inline int32_t ray_cast_f64_to_i32_null(double v) {
-    return (v != v) ? NULL_I32 : (int32_t)v;
+    if (v != v) return NULL_I32;
+    if (v >= (double)INT32_MAX) return INT32_MAX;
+    if (v <= (double)INT32_MIN) return INT32_MIN + 1;
+    return (int32_t)v;
 }
 
 static inline int16_t ray_cast_f64_to_i16_null(double v) {
-    return (v != v) ? NULL_I16 : (int16_t)v;
+    if (v != v) return NULL_I16;
+    if (v >= (double)INT16_MAX) return INT16_MAX;
+    if (v <= (double)INT16_MIN) return INT16_MIN + 1;
+    return (int16_t)v;
 }
 
 static inline uint8_t ray_cast_f64_to_u8_null(double v) {
-    return (v != v) ? 0 : (uint8_t)v;
+    if (v != v || v <= 0.0) return 0;
+    if (v >= (double)UINT8_MAX) return UINT8_MAX;
+    return (uint8_t)v;
 }
 
 static inline uint8_t ray_cast_f64_to_bool_null(double v) {
--- a/src/core/types.h
+++ b/src/core/types.h
@@ -42,4 +42,27 @@ extern const uint8_t ray_type_sizes[256];
 /* Element size for a given type tag */
 #define ray_elem_size(t)  (ray_type_sizes[(t)])
 
+static inline int64_t ray_cast_f64_to_i64_null(double v) {
+    if (v != v) return NULL_I64;
+    if (v >= (double)INT64_MAX) return INT64_MAX;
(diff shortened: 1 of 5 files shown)


src/core/types.h
```diff
@@ -42,4 +42,27 @@ extern const uint8_t ray_type_sizes[256];
 /* Element size for a given type tag */
 #define ray_elem_size(t)  (ray_type_sizes[(t)])
 
+static inline int64_t ray_cast_f64_to_i64_null(double v) {
```
singaraiona, line 45: Please also use this helper in `exec_elementwise_unary()`'s `in_type == RAY_F64 && out_type == RAY_I64` fallback arm. Its `OP_CAST` default at `src/ops/expr.c:2507` still does `dst[i] = (int64_t)src[i]`; when fusion bails and the source lane is `NULL_F64`, that casts NaN before the later `propagate_nulls()` pass can repair the payload. I forced the fallback path for the existing nullable F64-to-I64 test and ran with `-fsanitize=float-cast-overflow` plus `UBSAN_OPTIONS=halt_on_error=1`; it stopped at that line with `runtime error: nan is outside the range of representable values of type 'long int'`.

singaraiona (commented):
The new helpers cover the fused and narrow fallback casts, but the fallback F64-to-I64 cast still executes the original undefined conversion.

src/core/types.h
```diff
@@ -42,4 +42,27 @@ extern const uint8_t ray_type_sizes[256];
...
+    if (v != v) return NULL_I64;
+    if (v >= (double)INT64_MAX) return INT64_MAX;
+    if (v <= (double)INT64_MIN) return INT64_MIN + 1;
+    return (int64_t)v;
+}
+
+static inline int32_t ray_cast_f64_to_i32_null(double v) {
+    return (v != v) ? NULL_I32 : (int32_t)v;
```
singaraiona: [P2] Guard finite out-of-range values in the narrow helpers

The NaN branch removes the null-sentinel UB, but every other value still reaches the direct int32_t cast; the I16 and U8 helpers below have the same problem. A finite double outside the destination range is also undefined behavior under C11 6.3.1.4. I reproduced this with 1e300/-1e300 and -fsanitize=float-cast-overflow (the probe stops in ray_cast_f64_to_u8_null; I32/I16 fail likewise). Please add explicit range handling before these casts, analogous to the I64 helper and preserving the reserved signed null sentinels.

singaraiona (commented):
Follow-up review of the updated branch.

PR #341, opened 2026-07-23
perf: parallelize serial stages around group-by; unify binary-agg null semantics

ser-vasilich:
# perf: parallelize the serial stages around group-by; unify binary-agg null semantics

## Problem

On a 48-core machine the TAQ benchmark (medium NYSE dataset, quote = 296.6M rows) showed group-by queries scaling only **2.1–2.3×** from 1 to 48 cores. Profiling the full query pipeline showed the group-by kernel itself was fine after per-worker accumulation — the wall-clock was dominated by *serial* stages around it:

- binary aggregates (`pearson`/`cov`/`wavg`/`wsum`) were not DA-eligible → fell to slower paths;
- `OP_IF` projections went through `exec_if_selected` (serial per-row id-scaffolding) even when no WHERE was present;
- bitmap→index materialization in `exec_filter` / `sel_compact` was serial;
- the `where` builtin and `at col idx` gather were serial;
- `ray_pool_dispatch` was being fed chunk counts, so 1500+ chunks morselized into ~2 tasks.

## What changed

**`src/ops/group.c` — binary aggregates via the DA (dense-array) path.** Per-worker accumulators extended with co-moments (Σx, Σx², Σy, Σy², Σxy), finals in `emit_agg_columns`. Two merge bugs fixed along the way: Σx for integer x-columns is stored as double but was merged through the `int64` union member; the `da_merge_fn` path didn't merge the PAIR co-moment arrays at all. Eligibility is gated: y must be a numeric column of at least scan length, FP or non-nullable integer.

**`src/ops/group.c` — grouped binary-agg null semantics unified across every path.** The scalar reducers, the v2 engine and the new DA path all implement pair-skip (a null on either side voids the whole (x,y) pair); the legacy HT accumulators only null-checked the x side. Now the HT path pair-skips too: entry packing canonicalizes integer nulls (NaN in F64-packed slots, `NULL_I64` in int×int slots) and `accum_from_entry_nullable` skips the pair before `nn++`/Σx. Three latent HT bugs found and fixed in the process: nullable-int cells packed beside an FP partner read `(double)sentinel` as a huge finite value; both HT emitters divided pearson/cov moments by the group row count instead of the accumulated pair count; the all-null-group guards wrote `0.0` over the null sentinel, so an all-null group emitted `0.0` instead of a typed null.

**`src/ops/pivot.c` — `OP_IF`.** (1) The eager elementwise fill is pool-parallel (type-switched morsel fill; SYM LUT warm-up stays serial before dispatch; STR output stays serial). (2) Trivial-branch ifs (scan/const through alias/materialize) are rerouted from `exec_if_selected` to the eager path, gated by an out-type whitelist — `exec_if_selected` fires even without a WHERE clause and was the hidden per-row serial cost in projection-heavy group-bys.

**`src/ops/filter.c` — `exec_filter` and `sel_compact`.** Bitmap→`match_idx` build parallelized with 3-phase compaction (per-chunk popcount → serial exclusive prefix sum → per-chunk fill at disjoint offsets), one task per chunk via `ray_pool_dispatch_n`. Chunk count capped at 1024 (the initial task-ring capacity), so ring growth can never fail and silently drop tasks.

**`src/ops/builtins.c` / `src/lang/eval.c`.** `where` builtin parallel with the same 3-phase pattern (keeps the word-at-a-time zero-skip in the emit phase); `gather_by_idx` parallel; null-bit propagation stays serial.

**Safety gates.** All new dispatch sites go through one predicate (`ray_par_dispatch_ok`): live pool with background workers (at `-c 1` the pool exists with 0 workers), `len >= RAY_PARALLEL_THRESHOLD`, and a `ray_parallel_flag` reentrancy check (the task ring is single-producer). Serial fallback on any allocation failure.

## Results

48-core box, canonical `queryEngines.sh`, medium TAQ, min of 3 runs, **both binaries built from the same dev base** (`8f8033c`):

| Query | dev c48 (ms) | branch c48 (ms) | speedup | scaling 1→48 before | after |
|---|--:|--:|--:|--:|--:|
| Q22 (wavg by ex) | 3932 | **1443** | 2.72× | 2.1× | **5.6×** |
| Q23 | 3731 | **1405** | 2.66× | 2.2× | **5.7×** |
| Q24 (by sym,ex) | 6180 | **2293** | 2.70× | 2.3× | **5.9×** |
| Q25 | 5479 | **1770** | 3.10× | 2.2× | **6.7×** |

Side wins from the filter/where/if work: Q48 2.28×, Q42 1.69×, Q47 1.63×, Q7 1.49×, Q39 1.29×, Q8/Q16/Q17/Q19/Q41 1.2–1.3×.

**Single-core: no regressions** — branch ≤ dev at c1 across all 53 queries (e.g. Q22: 8052 vs 8216 ms), thanks to the worker-count gates.

## Testing

- `make test`: **3640/3640** on a clean ASan/UBSan build.
- New `test/rfl/agg/binary_group_null_paths.rfl`: 46 assertions covering wsum/wavg/pearson_corr/cov/scov on all three grouped routes (v2 / DA / legacy HT, both entry-packing lanes) against independently computed pair-skip truth, all four x/y type combinations with nulls on both sides, plus an all-pairs-null group asserting `wsum = 0.0` and typed nulls for the ratio/moment aggs on every route.
- `wavg`/`pearson` grouped results additionally verified against numpy to full precision, including FP columns with NaNs and integer x-columns.

## Known remaining gaps (follow-ups, out of scope here)

- Q22-25 still have headroom at high core counts: each op (`in`, `if`, gather, arithmetic) materializes its full result before the next starts; fusing the filter/projection expressions into one compaction pass would remove most of the remaining wall-clock.
- Window functions parallelize across partitions only (`window.c` Pass 3); low-cardinality partitioning caps their scaling.

Supersedes #338 (same branch; the audit-gate findings from that run are addressed in the last three commits).

https://claude.ai/code/session_01JZwH1gpAeLzDX2bxc4jz6s


singaraiona:
## Rayforce targeted audit passed

The required Rayforce audit gate passed on the latest run.

Workflow run: https://github.com/RayforceDB/rayforce/actions/runs/29999702106

PR #346, opened 2026-07-24
fix(ipc): preserve boxed data list args

belowzeroff:
## Summary
- mark nested list arguments in non-string IPC calls with a literal fallback
- preserve boxed data lists, including list-of-dicts payloads, when their head is not callable
- keep callable nested lists evaluated normally, preserving expression-style IPC messages

## Tests
- RAYFORCE_CORES=2 ./rayforce.test -f lang/eval/literal_fallback_preserves_boxed_data
- RAYFORCE_CORES=2 ./rayforce.test -f lang/eval/symbol_head_applies
- RAYFORCE_CORES=2 ./rayforce.test -f rfl/hof/eval_coverage2

Note: socket-backed IPC regression RAYFORCE_CORES=2 ./rayforce.test -f ipc/send_list_dict_arg_literal fails locally before exercising the change: ray_ipc_server_init returns RAY_ERR_IO, matching the existing local IPC/listen failures.

### DIFF
--- a/src/core/ipc.c
+++ b/src/core/ipc.c
@@ -59,6 +59,16 @@
 #include "lang/internal.h"
 #include "table/sym.h"
 
+static void mark_ipc_literal_fallbacks(ray_t* obj) {
+    if (!obj || RAY_IS_ERR(obj) || obj->type != RAY_LIST)
+        return;
+
+    obj->attrs |= RAY_EVAL_LITERAL_FALLBACK;
+    ray_t** elems = (ray_t**)ray_data(obj);
+    for (int64_t i = 0; i < ray_len(obj); i++)
+        mark_ipc_literal_fallbacks(elems[i]);
+}
+
 /* ===== Compression (delta + RLE) ===== */
 
 size_t ray_ipc_compress(const uint8_t* src, size_t len,
@@ -506,6 +516,12 @@ static ray_t* eval_payload_core(uint8_t* payload, size_t payload_len,
 
     ray_t* result = NULL;
     if (msg && !RAY_IS_ERR(msg)) {
+        if (msg->type == RAY_LIST) {
+            ray_t** elems = (ray_t**)ray_data(msg);
+            for (int64_t i = 1; i < ray_len(msg); i++)
+                mark_ipc_literal_fallbacks(elems[i]);
+        }
+
         /* Dispatch through `.ipc.on.sync` / `.ipc.on.async` hook if
          * installed; otherwise fall back to v1's inline-eval default.
          * The hook receives the raw deserialised payload — same shape
--- a/src/lang/eval.c
+++ b/src/lang/eval.c
@@ -3609,8 +3609,17 @@ ray_t* ray_eval(ray_t* obj) {
 
     /* List: evaluate first element, dispatch by type */
     ray_t** elems = (ray_t**)ray_data(obj);
+    int literal_fallback = (obj->attrs & RAY_EVAL_LITERAL_FALLBACK) != 0;
     ray_t* head = ray_eval(elems[0]);
-    if (RAY_IS_ERR(head)) { ret = head; goto out; }
+    if (RAY_IS_ERR(head)) {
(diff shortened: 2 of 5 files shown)


src/core/ipc.c
```diff
@@ -541,6 +551,11 @@ static ray_t* eval_payload_core(uint8_t* payload, size_t payload_len,
             }
             ray_release(msg);
         } else {
+            if (msg->type == RAY_LIST) {
```
singaraiona: The literal marker is applied only after the hook branch, so an installed `.ipc.on.sync` hook such as `(fn [m] (eval m))` receives the unmarked message and still rejects this same list-of-dicts argument. I reproduced this by extending `ipc/send_list_dict_arg_literal`: the original request passes without a hook, then fails after installing the eval hook. Could we mark non-string list arguments before `hook_lookup()` / `call_fn1()` so hook-based evaluation gets the same semantics?

src/lang/eval.c
```diff
@@ -3827,6 +3827,10 @@ ray_t* ray_eval(ray_t* obj) {
         default: {
             int8_t head_type = head->type;
             ray_release(head);
+            if (obj->attrs & RAY_EVAL_LITERAL_FALLBACK) {
```
singaraiona, line 3853: This fallback runs too late for boxed lists whose first element is a symbol. `ray_eval(elems[0])` or the subsequent symbol-resolution block returns a `name` error for `(list 'ipc_data_tag 1)` before reaching this switch, even though the head is non-callable. I reproduced it by adding that case to `lang/eval/literal_fallback_preserves_boxed_data`. Could the literal fallback also be honored when head evaluation or resolution yields an undefined symbol?

singaraiona (commented):
Two edge cases reproduced locally against the PR head:

src/core/ipc.c
belowzeroff: Addressed in 6a1821c1: list arguments are now marked before hook lookup/call, so `.ipc.on.sync` hooks using `(eval m)` see the same literal-fallback semantics as the default eval path. Added the hook case to `ipc/send_list_dict_arg_literal`.

src/lang/eval.c
belowzeroff: Addressed in 6a1821c1: literal fallback is now honored for undefined-symbol/name failures during head evaluation or symbol resolution, while non-name resolution errors still propagate. Added `(list 'ipc_data_tag 1)` coverage to `lang/eval/literal_fallback_preserves_boxed_data`.

src/lang/eval.c
```diff
@@ -3609,8 +3609,17 @@ ray_t* ray_eval(ray_t* obj) {
 
     /* List: evaluate first element, dispatch by type */
     ray_t** elems = (ray_t**)ray_data(obj);
+    int literal_fallback = (obj->attrs & RAY_EVAL_LITERAL_FALLBACK) != 0;
     ray_t* head = ray_eval(elems[0]);
-    if (RAY_IS_ERR(head)) { ret = head; goto out; }
+    if (RAY_IS_ERR(head)) {
+        const char* code = ray_err_code(head);
+        if (literal_fallback && code && strcmp(code, "name") == 0) {
```
singaraiona, line 3616: [P2] Restrict literal fallback to direct head-resolution failures

This catches any `name` error produced while evaluating `elems[0]`, including errors raised inside a callable head expression. I reproduced it by parsing `(((fn [] missing_name)) 1)`, setting the outer list's `RAY_EVAL_LITERAL_FALLBACK` flag, and calling `ray_eval`: the outer list is returned instead of the `name` error. That silently hides an execution failure in an IPC payload. Please only apply the fallback when the head itself is unresolved/non-callable, or otherwise distinguish head-resolution failure from errors raised during head execution.

singaraiona (commented):
Follow-up review of 6a1821c1: the two original findings are addressed. I found one new error-propagation edge case in the updated fallback logic.

PR #361, opened 2026-08-03
fix(builtins): reject malformed strings in TIMESTAMP cast

belowzeroff:
## Problem

`(as 'TIMESTAMP str)` silently accepted several malformed inputs instead of returning a `domain` error, producing valid-looking but wrong values — a data-integrity issue, since the corruption is invisible at the call site.

| Input | Before (wrong) | After |
|---|---|---|
| `"2024-01-02x01:02:03"` | `2024.01.02D00:00:00` (time dropped) | `domain` error |
| `"2024-01-02abc"` | `2024.01.02D00:00:00` | `domain` error |
| `"2024-01-02T25:02:03"` | `2024.01.03D01:02:03` (rolled to next day) | `domain` error |
| `"2024-01-02T12:99:03"` | `2024.01.02D13:39:03` | `domain` error |

Root cause: the time part was only parsed when the character at index 10 was `T`/`' '`/`D`; any other character silently skipped the entire time component and returned midnight. Out-of-range hour/minute/second values were normalized (carried into the following day) instead of rejected.

## Fix

- Require a recognized date/time separator (`T`, `' '`, or `D`) before a time part — otherwise `domain` error.
- Require the time to parse at least `HH:MM`.
- Reject out-of-range hour/minute/second components.

This matches the strict behavior of the `DATE` string cast.

## Compatibility

All previously accepted valid forms continue to round-trip: bare date (→ midnight), space/`T`/`D` separators, fractional seconds, and `Z` / `+HH:MM` / `-HHMM` timezone suffixes. Covered by the existing goldens in `ops/builtins_branch_cov.rfl`, which this PR extends with the rejection cases plus the bare-date and space-separator forms.

Full `make test` suite passes (0 failed) under the default ASan/UBSan build.

### DIFF
--- a/src/ops/builtins.c
+++ b/src/ops/builtins.c
@@ -1628,34 +1628,92 @@ ray_t* ray_cast_fn(ray_t* type_sym, ray_t* val) {
             int64_t days = val->i32;
             return ray_timestamp(days * 24LL * 60 * 60 * 1000000000LL);
         }
-        /* ISO string -> timestamp: "YYYY-MM-DD[T ]HH:MM:SS[.nnn...]" or "YYYY.MM.DDDHH:MM:SS.nnn..." */
+        /* ISO string -> timestamp.  Parse the whole value with a bounded
+         * cursor and require it to reach the end of the string.  A partial
+         * or trailing-garbage match must be REJECTED — otherwise malformed
+         * input is silently cast to a valid-but-wrong timestamp, the exact
+         * data-corruption class this parser guards against.  Grammar:
+         *   YYYY<sep>MM<sep>DD [ (T|' '|D) HH:MM[:SS][.frac] [Z|(+|-)HH[:]?MM] ]
+         * where <sep> is '-' or '.' and is consistent within the date. */
         if (val->type == -RAY_STR) {
             const char* sp = ray_str_ptr(val);
             size_t sl = ray_str_len(val);
-            if (sl < 10) return ray_error("domain", "as: cannot parse str as timestamp, too short, got %lld chars", (long long)sl);
-            int y, m, d, hh = 0, mm = 0, ss = 0;
-            long long frac = 0;
-            /* Try both formats: YYYY-MM-DD and YYYY.MM.DD */
-            int parsed = sscanf(sp, "%d-%d-%d", &y, &m, &d);
-            /* parse date: try YYYY-MM-DD then YYYY.MM.DD */
-            if (parsed != 3) {
-                parsed = sscanf(sp, "%d.%d.%d", &y, &m, &d);
-                /* YYYY.MM.DD format */
-            }
-            if (parsed != 3) return ray_error("domain", "as: cannot parse str as timestamp, expected YYYY-MM-DD or YYYY.MM.DD");
-            /* Parse optional time part */
-            if (sl > 10 && (sp[10] == 'T' || sp[10] == ' ' || sp[10] == 'D')) {
-                sscanf(sp + 11, "%d:%d:%d", &hh, &mm, &ss);
-                /* Parse fractional seconds */
-                const char* dot = memchr(sp + 11, '.', sl - 11);
-                if (dot) {
-                    dot++;
+            size_t i = 0;
+            int y = 0, m = 0, d = 0, hh = 0, mm = 0, ss = 0;
+            long long frac = 0, tz_ns = 0;
+
+            /* Read exactly `w` digits (0-9) into `out`, or fail. */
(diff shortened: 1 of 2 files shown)


singaraiona (changes requested):
Blocking correctness issue: the timestamp parser still accepts partial matches and trailing garbage.

At `src/ops/builtins.c:1655`, `nr >= 2` only proves that `sscanf` parsed the hour and minute; it does not prove that the complete input is valid or consumed. Against commit `12e6c761`, I reproduced successful casts for:

- `2024-01-02T12:34junk`
- `2024-01-02T12:34:`
- `2024-01-02T12:34:03junk`
- `2024-01-02T12:34:03Zjunk`
- `2024-01-02T12:34:03+99:99` (normalizes to a different date)

This preserves the silent data-corruption class the PR is intended to eliminate. Please parse with a bounded cursor through the exact `HH:MM[:SS][.fraction][Z|±HH[:]?MM]` grammar, require the cursor to reach the string length, validate timezone ranges, and add rejection tests for partial/trailing components.

The full suite still passes (`3652/3652`), so the current tests do not catch these cases.

belowzeroff:
Thanks — you're right, `nr >= 2` only proved a prefix matched and left the trailing/partial/out-of-range cases open. Fixed in `15e6b442` by replacing the `sscanf` prefix matching with a **bounded cursor** over the full grammar:

```
YYYY<sep>MM<sep>DD [ (T|' '|D) HH:MM[:SS][.frac] [Z|(+|-)HH[:]?MM] ]
```

- `<sep>` is `-` or `.` and must be consistent within the date; each field is a fixed digit width.
- The cursor **must reach the end of the string** — no trailing characters are ignored.
- Date, time, and timezone components are all range-checked.

All five cases you reproduced now return a `domain` error:

| Input | Result |
|---|---|
| `2024-01-02T12:34junk` | domain (trailing garbage) |
| `2024-01-02T12:34:` | domain (dangling `:`) |
| `2024-01-02T12:34:03junk` | domain (trailing garbage) |
| `2024-01-02T12:34:03Zjunk` | domain (trailing after `Z`) |
| `2024-01-02T12:34:03+99:99` | domain (tz out of range) |

Also now rejected: `12:34:03.` (dangling fraction), inconsistent date separators (`2024-01.02...`), and non-fixed-width fields (`2024-1-2...`).

Every previously accepted valid form still round-trips (bare date, space/`T`/`D` separators, fractional seconds of any length, and `Z` / `+HH:MM` / `-HHMM` offsets) — verified against the existing goldens plus the new rejection cases added to `ops/builtins_branch_cov.rfl`. Full `make test` passes (3648/3649, 1 skipped, 0 failed) under ASan/UBSan.

singaraiona (changes requested):
Re-reviewing head `2740013f`: the bounded cursor fixes the original five reproductions, but three blocking normalization/overflow cases remain.

1. `src/ops/builtins.c:1701-1703` makes timezone minutes optional after parsing the hour. Both `2024-01-02T12:34:03+12` and `2024-01-02T12:34:03+12:` are accepted as `+12:00`, although the documented grammar requires `HH[:]?MM`. The two minute digits must be mandatory after the optional colon.

2. `src/ops/builtins.c:1664` only checks `d <= 31`, not the selected month/leap year. The new parser accepts and normalizes `2024-02-31` to March 2, `2023-02-29` to March 1, and `2024-04-31` to May 1. Validate against the actual days in the parsed month.

3. `src/ops/builtins.c:1728` multiplies an unrestricted four-digit year's day count by nanoseconds per day without checked arithmetic or a representable-range guard. Under the debug UBSan build, `(as 'TIMESTAMP "9999-12-31")` reports signed integer overflow and returns a bogus 1816 timestamp. Reject dates/timezone-adjusted results outside the `int64_t` nanosecond range before performing the arithmetic.

Please add regressions for these cases. I am leaving the change request in place pending another revision.

belowzeroff:
Addressed the re-review findings in 8dd20bd7.

- Timezone offsets now require the minute field after the hour, so `+12` and `+12:` are rejected instead of being normalized to `+12:00`.
- Date validation now checks the day against the parsed month and leap year, so invalid calendar dates like `2024-02-31`, `2023-02-29`, and `2024-04-31` are rejected before normalization.
- Timestamp nanosecond construction now checks the int64 range before day multiplication, time-of-day addition, and timezone adjustment, so out-of-range values like `9999-12-31` return `domain` without UBSan overflow.

Added regressions for all of the above plus a valid leap-day control case. Verified locally with `make test TEST_CORES=2`: `3648 of 3649 passed (1 skipped, 0 failed)` under ASan/UBSan.

PR #386, opened 2026-08-07
fix(temporal): guard int64 overflow in DATE/TIMESTAMP decomposition

belowzeroff:
## Problem

The DATE/TIMESTAMP → microseconds conversion shared by the temporal **extract** and **truncate** paths overflowed `int64` on extreme inputs (UBSan):

```
(dd (as 'DATE 2147483647))
src/ops/temporal.c:105: runtime error: signed integer overflow:
  2147483647 * 86400000000 cannot be represented in type 'long long'

(mm (as 'TIMESTAMP -9223372036854775807))
src/ops/temporal.c:109: runtime error: signed integer overflow:
  9223372036854775807 + 999 cannot be represented in type 'long long'
```

- **DATE** is `int32` days, so an extreme value × µs-per-day overflows.
- The **TIMESTAMP** ns→µs floor negated the input (`-((-raw)+999)/1000`), so a value within 999 of `INT64_MIN` overflowed.

Both appear in **all four** decomposition kernels: the standalone `ray_temporal_extract` / `ray_temporal_truncate` and the DAG `exec_extract` / `exec_date_trunc` morsel kernels (reachable via `(dd d)` and via dotted-path `col.yyyy` / `col.date` in `select`).

## Fix

- Do the TIMESTAMP ns→µs floor with **truncate-then-adjust** so it never negates — overflow-free and exact at `INT64_MIN`.
- A DATE so extreme its µs value is not representable (hundreds of millennia out) decodes to a **null** instead of reading overflow garbage, consistent with how these kernels already treat a null input.
- Truncate additionally nulls a result whose bucketed µs would overflow the `int64` nanosecond output.

## Behaviour

| Expression | Before | After |
|---|---|---|
| `(yyyy (as 'DATE 2147483647))` | UB | `0Nl` (null) |
| `(mm (as 'TIMESTAMP -9223372036854775807))` | UB | `9` |
| `(yyyy 2024.03.15)` | `2024` | `2024` (unchanged) |
| `select d.yyyy` over `[0 366 2147483647]` | UB | `[2000 2001 0Nl]` |
| `select d.date` over the same | UB | `[… … 0Np]` |

Ordinary values are unaffected. New coverage in `temporal/extract_trunc_overflow.rfl` exercises the standalone and both DAG kernels with extreme DATE, minimum-edge TIMESTAMP, and ordinary values. Full `make test` passes (3660/3661, 1 skipped, 0 failed) under the default ASan/UBSan build.

### DIFF
--- a/src/ops/temporal.c
+++ b/src/ops/temporal.c
@@ -134,20 +134,31 @@ static inline bool rte_extract_elem(int8_t t, int64_t raw, int field, int64_t* o
 }
 
 /* Truncate a raw temporal slot to a TIMESTAMP-ns bucket.  Returns false when
- * the source or the bucketed result is outside the int64 nanosecond range
+ * the source, or the bucketed result, is outside the int64 nanosecond range
  * (caller emits null): a value hundreds of millennia out cannot be a
- * TIMESTAMP.  Keeps one bucket of headroom because the floor step can
- * subtract up to `bucket` µs before the ×1000 ns conversion.  The headroom
- * is applied unconditionally (not only when a floor actually happens), so up
- * to one bucket of representable values at the very low edge round to null —
- * an immaterial band ~292 millennia before 2000, not worth an extra branch
- * in this hot kernel. */
+ * TIMESTAMP.  The floor uses overflow-safe arithmetic and the ACTUAL bucketed
+ * result is range-checked (mirroring exec_date_trunc), so a value whose
+ * floored bucket is still representable — e.g. the low day boundary
+ * 1707.09.23 — round-trips instead of being rejected by a pre-floor headroom
+ * that disagreed with the DAG path. */
 static inline bool rte_trunc_elem(int8_t t, int64_t raw, int64_t bucket, int64_t* out_ns) {
     int64_t us;
     if (!rte_to_us_ck(t, raw, &us)) return false;
-    if (us > INT64_MAX / 1000LL || us < INT64_MIN / 1000LL + bucket) return false;
+    /* Floor `us` to the bucket boundary (toward -inf).  `us - r` truncates
+     * toward zero (magnitude only shrinks, never overflows); a negative
+     * remainder needs one more bucket subtracted to floor toward -inf — the
+     * one step that can underflow (a DATE reaches `us` across the full µs
+     * range), so guard it. */
     int64_t r = us % bucket;
-    int64_t out_us = us - r - (r < 0 ? bucket : 0);
+    int64_t out_us = us - r;
+    if (r < 0) {
+        if (out_us < INT64_MIN + bucket) return false;
+        out_us -= bucket;
+    }
+    /* Range-check the ACTUAL floored result against the int64 nanosecond
+     * domain (out_us × 1000), matching exec_date_trunc — not a conservative
+     * pre-floor headroom, which rejected representable boundary values. */
+    if (out_us > INT64_MAX / 1000LL || out_us < INT64_MIN / 1000LL) return false;
     *out_ns = out_us * 1000LL;
     return true;
 }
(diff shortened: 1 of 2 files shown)


belowzeroff:
Addressed the `rayforce-audit` findings in `ae215c59`.

**Blocking (1) — incomplete truncate guard.** Correct: the `exec_date_trunc` `YEAR`/`MONTH` arms re-multiply `days_from_civil(...) * µs-per-day` on a day count floored *down* to the period start (up to a year beyond the input), so a DATE that cleared the µs bound (`INT64_MAX / µs-per-day`) still overflowed — `d.year` of `(as 'DATE -106751991)` tripped it (`-106752347 * 86400000000`). Tightened the IN32 DATE guard to the int64-**nanosecond** representable day range (`INT64_MAX / 1000 / DT_USEC_PER_DAY`), matching the sibling `rte_trunc_elem`; the re-multiply and the final ×1000 now both stay in range.

**(2) — vacuous assertion.** Right, `-2147483648 == NULL_I32`, so it exited at the null-input early-out. Replaced with `-200000000` (a real non-null negative that exercises the guard's negative branch).

**(3) — untested standalone truncate / overstated header.** Added `(date <TIMESTAMP>)` / `(date <DATE>)` cases (ordinary → day-truncated, out-of-ns-range → null) and the DAG `d.year` / `d.month` re-multiply path with a large-magnitude negative DATE. Corrected the header to describe exactly what is covered.

**(4) — conservative low-edge guard.** Kept the unconditional one-bucket headroom (avoids a branch in the hot kernel) and documented that it rounds an immaterial ~1-bucket band at the extreme low edge (~292 millennia before 2000) to null.

Full `make test` still green (3660/3661, 1 skipped, 0 failed) under ASan/UBSan; the `d.year (as 'DATE -106751991)` repro no longer reports UB.

src/ops/temporal.c
```diff
@@ -101,16 +101,56 @@ static int64_t rte_extract_one(int64_t us, int field) {
...
+ * TIMESTAMP.  Keeps one bucket of headroom because the floor step can
+ * subtract up to `bucket` µs before the ×1000 ns conversion.  The headroom
+ * is applied unconditionally (not only when a floor actually happens), so up
+ * to one bucket of representable values at the very low edge round to null —
+ * an immaterial band ~292 millennia before 2000, not worth an extra branch
+ * in this hot kernel. */
+static inline bool rte_trunc_elem(int8_t t, int64_t raw, int64_t bucket, int64_t* out_ns) {
+    int64_t us;
+    if (!rte_to_us_ck(t, raw, &us)) return false;
+    if (us > INT64_MAX / 1000LL || us < INT64_MIN / 1000LL + bucket) return false;
```
singaraiona: [P2] This unconditional bucket headroom rejects valid timestamps before checking the actual truncated result. For example, `(date (as 'TIMESTAMP -9223286400000000000))` returns `0Np`, although that value is already the representable day boundary `1707.09.23D00:00:00.000000000`. The equivalent DAG path (`select ts.date`) returns that timestamp, so the two public paths disagree. Please compute the bucketed value with overflow-safe arithmetic, range-check the actual result, and add this boundary as a regression.

singaraiona (changes requested):
Found one correctness issue at the lower timestamp boundary. The overflow fixes otherwise look sound, CI is green, and the full local ASan/UBSan suite passed (3664/3664).

belowzeroff:
Addressed the review in `686abd1f`.

`rte_trunc_elem` now floors with overflow-safe arithmetic (truncate toward zero, then a single guarded toward-−∞ bucket subtraction) and range-checks the **actual** bucketed result against the int64-ns domain — mirroring `exec_date_trunc` — instead of the unconditional pre-floor `+ bucket` headroom.

`(date (as 'TIMESTAMP -9223286400000000000))` now returns `1707.09.23D00:00:00.000000000`, matching the DAG `ts.date` path, so the two public truncate paths agree at the boundary. The extreme-min case (`-9223372036854775807`, whose floored day boundary genuinely spills past the ns range) still nulls.

Added that boundary — standalone `(date …)` **and** DAG `ts.date` — as a regression in `extract_trunc_overflow.rfl`. Full `make test` green (3660/3661, 1 skipped, 0 failed) under ASan/UBSan; the repro no longer disagrees across paths.

PR #388, opened 2026-08-11
fix(arith): wrap scalar integer add/sub/mul on overflow instead of UB

belowzeroff:
## Problem

`ray_add_fn` / `ray_sub_fn` / `ray_mul_fn` computed `as_i64(a) OP as_i64(b)` directly, so an i64 result that overflowed was signed-integer-overflow undefined behavior:

```
(+ 9223372036854775807 1)   src/ops/arith.c:141  signed integer overflow
(- 0 -9223372036854775808)  src/ops/arith.c:223  signed integer overflow
(* 9223372036854775807 2)   src/ops/arith.c:257  signed integer overflow
```

The wrapped value the code relied on — `INT64_MIN` for `MAX+1`, which is the i64 null sentinel — was already what it returned, so the behavior is correct in practice; only the arithmetic itself is undefined. The vector arithmetic kernel (`src/ops/expr.c` `OP_ADD`/`SUB`/`MUL`) already wraps via `uint64`.

## Fix

Compute the integer result with `uint64` wraparound (defined), matching the vector kernel. Results are unchanged:

| Expression | Before | After |
|---|---|---|
| `(+ 9223372036854775807 1)` | UB (→ `0Nl`) | `0Nl` |
| `(- -9223372036854775807 2)` | UB | `9223372036854775807` |
| `(* 9223372036854775807 2)` | UB | `-2` |
| `(+ 2 3)` | `5` | `5` |

Narrow-type (i16/i32) results are unaffected — their i64 intermediate never overflows and `make_typed_int` still narrows them. `neg`/`abs`/`div`/`mod` already guard their overflow cases and are left as-is.

## Tests

Adds `test/rfl/arith/overflow_wrap.rfl`. Full `make test` passes (3663/3664, 1 skipped, 0 failed) under the default ASan/UBSan build.

### DIFF
--- a/src/ops/arith.c
+++ b/src/ops/arith.c
@@ -138,7 +138,12 @@ ray_t* ray_add_fn(ray_t* a, ray_t* b) {
     if (both_f32_atoms(a, b)) return ray_f32((float)(as_f64(a) + as_f64(b)));
     if (is_float_op(a, b)) return make_f64(as_f64(a) + as_f64(b));
     int8_t rt = promote_int_type(a, b);
-    return make_typed_int(rt, as_i64(a) + as_i64(b));
+    /* Wrap on overflow (defined) rather than signed-overflow UB, matching the
+     * vector arithmetic kernel (src/ops/expr.c OP_ADD/SUB/MUL).  The wrapped
+     * i64 (INT64_MIN on MAX+1) is the null sentinel, so results are unchanged
+     * — only the undefined arithmetic is removed. */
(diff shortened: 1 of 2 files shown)


singaraiona (approved):
Reviewed the scalar integer add/sub/mul wrap semantics and regression coverage. The implementation matches the existing vector kernel. Verified with the debug ASan/UBSan build and full make test: 3667/3667 passed.

PR #389, opened 2026-08-11
fix(builtins): reject malformed strings in TIME cast

belowzeroff:
## Problem

`(as 'TIME str)` parsed with an unanchored `sscanf` and never range-checked the minute/second fields, so it silently accepted malformed input and produced a wrong-but-valid value — the same data-corruption class the DATE (#382) and TIMESTAMP (#361) string casts were hardened against, but the TIME cast was left on the old path:

| Input | Before | After |
|---|---|---|
| `"12:34:56junk"` | `12:34:56.000` (garbage ignored) | `domain` error |
| `"12:34:56.789junk"` | `12:34:56.789` | `domain` error |
| `"25:99:99"` | `26:40:39.000` (min/sec out of range, normalized) | `domain` error |
| `"12x34x56"` | `domain` | `domain` |

## Fix

Parse `[-]HH:MM[:SS][.fff]` with a bounded cursor that must consume the whole string, validating that minutes and seconds are `0-59`.

TIME is a **signed duration** — its ms-of-day may exceed a day and go negative (see `.csv.read` round-tripping, #379) — so the hour field stays variable width and unbounded, `HH:MM` (no seconds) is accepted, and a bare trailing `.` is accepted as `.000`. The result is range-checked against the int32 millisecond domain. The string-**vector** cast routes through this same atom path, so it is fixed too.

Preserved (still accepted): `12:30:45`, `12:30:45.123`, `20:00:00.`, `25:00:00` (>24h duration), `-5:00:00` (negative), `12:34`.

## Tests

Extends the TIME-cast coverage in `type/as.rfl` with the duration, negative, and rejection cases. Full `make test` passes (3662/3663, 1 skipped, 0 failed) under the default ASan/UBSan build.

### DIFF
--- a/src/ops/builtins.c
+++ b/src/ops/builtins.c
@@ -1627,21 +1627,60 @@ ray_t* ray_cast_fn(ray_t* type_sym, ray_t* val) {
              * matching wall-clock semantics. */
             return ray_time((int64_t)(ts_ns_in_day(val->i64) / 1000000LL));
         if (val->type == -RAY_STR) {
-            /* Parse "HH:MM:SS[.mmm]" */
+            /* Parse "[-]HH:MM[:SS][.fff]" with a bounded cursor that must
+             * consume the whole string.  TIME is a signed duration — ms-of-day
+             * can exceed a day and go negative (see .csv.read round-tripping) —
+             * so the hour field is variable width and unbounded, but minutes
+             * and seconds are 0-59.  The old sscanf path accepted trailing
+             * garbage ("12:34:56junk") and out-of-range fields ("25:99:99" →
+             * 26:40:39), silently corrupting the value; mirror the strict DATE
+             * / TIMESTAMP string casts. */
             const char* sp = ray_str_ptr(val);
-            int th = 0, tm = 0, ts = 0, tms = 0;
-            int nr = sscanf(sp, "%d:%d:%d", &th, &tm, &ts);
-            if (nr < 2) return ray_error("domain", "as: cannot parse str as time, expected HH:MM:SS[.mmm]");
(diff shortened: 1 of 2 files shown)


src/ops/builtins.c
```diff
@@ -1627,21 +1627,60 @@ ray_t* ray_cast_fn(ray_t* type_sym, ray_t* val) {
...
+                    i++;
+                }
                 tms = (int)strtol(mbuf, NULL, 10);
             }
-            int32_t ms = (int32_t)th * 3600000 + (int32_t)tm * 60000 + (int32_t)ts * 1000 + tms;
-            return ray_time((int64_t)ms);
+            if (i != sl)
+                return ray_error("domain", "as: cannot parse str as time, unexpected trailing characters");
+            int64_t ms = sign * (hh * 3600000LL + (int64_t)mm * 60000LL + (int64_t)ss * 1000LL + tms);
+            if (ms > INT32_MAX || ms < INT32_MIN)
```
singaraiona, line 1681: [P2] Reject the reserved TIME null sentinel

`INT32_MIN` is not a usable TIME payload: it is `NULL_I32`, and TIME null checks render it as `0Nt`. This condition admits exactly `"-596:31:23.648"` (-2,147,483,648 ms), so a non-null string silently casts to null; the string-vector path will likewise mark that cell null. I reproduced this on the PR head, while `"-596:31:23.647"` remains a valid duration. Please reject `ms == NULL_I32` (for example, use `ms <= INT32_MIN`) and add boundary coverage for both values.

singaraiona (changes requested):
Found one boundary correctness issue in the new TIME range check.

PR #400, opened 2026-08-14
fix(query): reject STR equality keys in asof-join and window-join

belowzeroff:
## Problem

`asof-join`, `window-join`, and `window-join1` accept string (`STR`) equality-key columns and then return **silently wrong** results — no error, corrupt data.

Root cause: all three read equality-key cells through `read_col_i64`, which has no `RAY_STR` case, so a `STR` column falls into the byte-wide default (`((uint8_t*)data)[row]`).

- **asof-join** (`asof_eq_lread`) mismatches outright — a key that should match nulls out:

  ```
  ;; identical data, only the key TYPE differs
  (asof-join [Sym Time] tr qs)  with SYM Sym -> Bid [99 99]   correct
  (asof-join [Sym Time] tr qs)  with STR Sym -> Bid [99 0Nl]  WRONG
  ```

- **window-join / window-join1** read one byte per row when they sort and probe the right side. Small key sets coincide by luck, but once there are more distinct keys than fit in a byte the collisions cross-contaminate groups:

  ```
  ;; 300 distinct keys, each with its own quote; per-key minBid must equal the key
  (sum (- minBid (til 300)))  with SYM keys ->      0   correct
  (sum (- minBid (til 300)))  with STR keys -> -41605   WRONG
  ```

## Fix

The base `inner-join` / `left-join` / `anti-join` use a **separate STR-aware kernel** (`ray_join` / `ray_antijoin`) and stay correct even with 300 distinct STR keys (verified), so this is scoped to the two window/asof kernels only.

Decline a `STR` equality key on either side with `nyi` instead of returning corrupt data (parted columns unwrapped to their base type first, matching the base join). One guard in `ray_asof_join_core` covers both asof call paths (plain + parted/day); one in `window_join_impl` covers both `window-join` and `window-join1`.

`nyi` (rather than adding STR support to these kernels) is the conservative bug fix: it turns silent corruption into an explicit, catchable error. Real STR eq-key support in the window/asof kernels can be a follow-up.

## Tests

`test/rfl/join/str_key_nyi.rfl`:
- `STR` eq-key -> `nyi` for `asof-join` (both operand orders), `window-join`, `window-join1`;
- `SYM` eq-keys keep working and stay order-independent on the right side;
- base `inner-join` / `anti-join` with `STR` keys must **not** be over-rejected (regression guard).

`make test`: 3673 of 3674 passed (1 skipped, 0 failed).

### DIFF
--- a/src/ops/query.c
+++ b/src/ops/query.c
@@ -14876,6 +14876,21 @@ static ray_t* window_join_impl(ray_t** args, int64_t n, int mode) {
                 for (int i = 0; i < 4; i++) ray_release(eargs[i]);
                 return ray_error("domain", "window-join: equality key column not found in both tables");
             }
+            /* STR equality-key columns are silently mismatched: window-join
+             * sorts and probes eq cells through read_col_i64, which has no
+             * RAY_STR case (it reads one raw byte per row), so once there are
+             * enough distinct keys the byte collisions cross-contaminate
+             * groups and the aggregates are WRONG.  Decline both sides — the
+             * base equi-join has a STR-aware kernel, these window kernels do
+             * not. */
+            int8_t lct = left_eq[e]->type, rct = right_eq[e]->type;
+            if (RAY_IS_PARTED(lct)) lct = (int8_t)RAY_PARTED_BASETYPE(lct);
+            if (RAY_IS_PARTED(rct)) rct = (int8_t)RAY_PARTED_BASETYPE(rct);
+            if (lct == RAY_STR || rct == RAY_STR) {
+                scratch_free(eq_hdr);
+                for (int i = 0; i < 4; i++) ray_release(eargs[i]);
+                return ray_error("nyi", "window-join: string equality key columns are not supported");
+            }
         }
 
         /* Parse every (name, (op src)) pair from the agg dict.  The dict's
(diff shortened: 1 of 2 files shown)


singaraiona:
Blocking: the new `STR` guard does not cover the supported 4-argument compatibility form of `window-join`.

The guard is inside the `n >= 5` Rayforce-convention branch of `window_join_impl`, so `(window-join left right [eq-keys] time)` bypasses it and falls through to `ray_asof_join` / `exec_window_join`. I reproduced this on PR head `da97945` with the same `STR` fixture:

```rayfall
(at (window-join trades quotes [Sym] 'Time) 'Bid)
```

It still returns `[99 0Nl]` rather than `nyi`, so silent corruption remains. The existing `lang/eval/window_join` test confirms this call shape is supported.

Please add the validation to the legacy fallthrough as well (or, preferably, reject `STR` after equality columns are resolved in the shared as-of executor), and add regressions for the 4-argument `window-join` and `window-join1` paths. The new test currently exercises only the 5-argument window form, which is why the full suite still passes.

For context, the previous related fixes covered adaptive-width/cross-domain `SYM`; they did not add `STR` descriptor handling to `read_col_i64`.

PR #424, opened 2026-08-22
fix(collection,query): harden take/drop count-overflow family

belowzeroff:
Follow-up to the review on #419 (whose INT64_MIN guard already landed via #409). That review found the guard stopped short of its own premise — sibling count paths kept the same signed-overflow UB, plus a pre-existing range-take overflow and a take/drop null-policy asymmetry. This PR closes all four, verified under UBSan against `dev`.

## 1. `select take:` INT64_MIN negation UB (query.c)
`select … take: -9223372036854775808` fed the count through `-atom_n` / `-n_take` in the no-sort, `apply_sort_take`, and DAG tail-pushdown paths — signed-overflow UB (`query.c:635/870/9941`) that also cascaded into `nrows - n` in the TAIL kernel (`exec.c:3022`). These paths build the range themselves, so the `ray_take_fn` guard never saw them. Now each site rejects the null-sentinel count with a `type` error before the negation; the upstream guards also stop the `exec.c` cascade.

## 2. Range-take `end = start + amount` overflow (collection.c) — memory-unsafe
`(take [1 2 3] [1 9223372036854775807])` overflowed `end` in the string / vector / dict / list branches → negative count → `memcpy` with a garbage size (crash / corruption on release; UBSan trap + `oom` on debug). The clamped end is now computed without ever forming an overflowing `start + amount` (`min(amount, len - start)`); the dict branch keeps its `end < start` guard since it doesn't early-return on `start >= len`.

## 3. take/drop null-sentinel asymmetry — unified
INT64_MIN is the i64 null sentinel (`0N`). `(take xs 0N)` erroring while `(drop xs 0N)` silently emptied was a footgun. Both verbs now reject it as a `type` error, matching `ts_window_arg`. (Updates the rfl test that pinned drop→empty.)

## 4. Guard cleanup (collection.c)
`as_i64` on an `-RAY_F32` atom bit-reads the storage, so an f32 count whose bits alias INT64_MIN would trip the guard with a misleading message — f32 is now rejected up front alongside f64. Dropped the redundant `ray_is_atom` check (`is_numeric` already implies an atom type code).

## Testing
- `make test` — full suite green (3701/3702 pass, 1 pre-existing skip, 0 failed), ASan/UBSan clean.
- Coverage added: range-take overflow clamps and the unified type-error policy in `collection/take.rfl` and `collection/drop_cut_rotate_cross.rfl`; the three `select take:` paths (pushdown / sorted / grouped) plus valid negative takes in `table/select.rfl`.

Closes the review items from #419; #419 itself can be closed as superseded (its guard is already in `dev` via #409).

### DIFF
--- a/src/ops/collection.c
+++ b/src/ops/collection.c
@@ -1629,7 +1629,7 @@ ray_t* ray_take_fn(ray_t* vec, ray_t* n_obj) {
      * floats up front: as_i64(f64) reads the bit pattern and would cause
      * e.g. (take 1.0 2.0) to attempt a 4.6-quintillion-element allocation
      * and surface as "oom" — misleading for what is really a type error. */
-    if (ray_is_atom(n_obj) && n_obj->type == -RAY_F64)
+    if (ray_is_atom(n_obj) && (n_obj->type == -RAY_F64 || n_obj->type == -RAY_F32))
         return ray_error("type", "take: count must be an integer, got %s", ray_type_name(n_obj->type));
     /* Range take: (take collection [start amount]) — slice from start for amount elements */
     if (ray_is_vec(n_obj) && n_obj->type == RAY_I64 && ray_len(n_obj) == 2) {
@@ -1662,8 +1662,9 @@ ray_t* ray_take_fn(ray_t* vec, ray_t* n_obj) {
             if (start < 0) start = slen + start;
             if (start < 0) start = 0;
             if (start >= slen) return ray_str("", 0);
-            int64_t end = start + amount;
-            if (end > slen) end = slen;
+            /* amount can be up to INT64_MAX, so `start + amount` overflows;
+             * take min(amount, slen - start) instead (0 <= start < slen). */
+            int64_t end = (amount < slen - start) ? start + amount : slen;
             return ray_str(s + start, (size_t)(end - start));
         }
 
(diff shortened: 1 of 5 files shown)


singaraiona (approved):
Reviewed with a full verification pass (ASan+UBSan build of the branch, all repro inputs re-run). All four items check out — this is approvable as-is.

**1. `select take:` negation guards — correct and complete.** Guards precede all three negations (query.c:629→:641 no-sort, :872→:881 sorted, :9943→:9959 DAG pushdown), and I grepped query.c exhaustively — those are the only take-count negations; the fused top-k paths accept positive atoms only, and exec.c's OP_TAIL `nrows - n` needed no change since its only producer is the now-guarded :9959. Error surface uniform, cleanup on the new error paths matches the neighbors. Runtime-verified all three error `type` while `take: -2` / `asc: v take: -2` tail semantics stay intact.

**2. Range-take clamp — this is the right form.** I specifically checked whether the clamp avoids executing the overflowing add (a post-hoc `if (end < start)` would itself be UB-reliant): `end = (amount < len - start) ? start + amount : len;` only performs the add when it provably fits, and the preconditions hold at all four sites (`amount >= 0` enforced at :1639; string/vec/list early-return on `start >= len`; the dict branch's negative `len - start` case falls to `end = len` with the retained `end < start` guard yielding count 0, as the comment explains; `len - start` itself can't overflow). Verified under UBSan: `(take [1 2 3] [1 9223372036854775807])` → `[2 3]`, string/list/dict variants clean, no traps.

**3. Unified null policy — correct.** Both verbs now reject INT64_MIN/`0N` with the same `type`-class error and wording; the rfl test pinning drop→empty is updated; `ray_drop_fn` has no other C caller, so nothing relied on the old clamp. Large-but-representable negative drops still clamp (`-9223372036854775807` → `[]`, `-2` → `[1 2 3]`).

**4. F32 rejection + guard cleanup — correct.** `-RAY_F32` rejected at :1632, well before the first `as_i64` use, which also makes dropping `ray_is_atom` from the guard safe (`is_numeric` only matches atom codes, and the bit-pattern fallback can no longer misfire).

Full suite on the branch: 3705/3705 under ASan+UBSan, and the new assertions pin every behavior above.

One optional nit: the take.rfl comment says F32 atoms have no rfl surface so only F64 is pinned — an F32 atom is actually producible via `.csv.read [F32]` + `at` (the recipe cov5.rfl already uses), and I confirmed `(take [1 2 3] <f32-atom>)` errors correctly, so the F32 case could be pinned too if you want the coverage.

PR #425, opened 2026-08-24
fix(join): compare GUID keys by full cell bytes

belowzeroff:
## Summary
- hash GUID join keys from the full 16-byte cell instead of scalar key reads
- compare GUID join keys with full-cell byte equality
- add a regression covering left, inner, full, and anti joins with GUID keys from a distinct right buffer

## Tests
- make test
- ./rayforce.test -f rfl/join/guid_key


singaraiona (changes requested):
Reviewed with a verification pass on a fresh build of the branch (all repros below executed against it). The change itself is correct: `ray_data` is slice-safe with 16-byte GUID scaling, the null check precedes both new branches, hash and eq are mutually consistent, and `ray_hash_bytes(...,16)` + `memcmp(...,16)` matches the codebase-wide GUID idiom. All equi-join paths (left/inner/full/anti/semi, radix and chained-HT) funnel through the two patched functions, and the STR/I64 fast paths are type-gated away from GUID — the equi-join family is genuinely fixed.

But the fix stops at the equi-join kernel, and the same silent-wrong-result class survives one kernel over, in the same file:

**1. asof-join GUID eq keys — silently wrong.** The asof kernel reads equality cells through `read_col_i64` (join.c:2236 and the staging sites at ~2848/3005/3019), whose default arm returns the single byte at `data[row]` — wrong width and wrong offset — and the planner guard at `query.c:15903` declines only RAY_STR. Executed repro: `(asof-join [k t] L R)` with a 2-row GUID key column and matching earlier right rows returns `b = [0Nl 0Nl]` instead of `[20 10]` — every match dropped.

**2. window-join / window-join1 — same gap, corrupts aggregates.** Guards at `query.c:15204` and the legacy fall-through at `:15776` also decline only STR. Executed repro: window-join over GUID keys where the trade's key has in-window quotes (expected minBid 100) returns `[0Nl 0Nl]`. This exactly mirrors the STR bug that `str_key_nyi.rfl` documents and guards against — GUID needs either the real fix or, at minimum, the same `nyi` guard extended to GUID at all three entry points so the failure is loud.

**3. F64 eq keys — identical mechanism, worth catching in the same sweep.** `read_col_i64` has no F64 arm either (equi-join handles F64 explicitly; the asof/window guards don't). Executed repro: asof-join with F64 keys `[1.5 2.5]` vs right `[2.5 1.5]` returns `b = [20 20]` instead of `[20 10]` — row 2 matches the wrong key's quote via a one-byte compare.

The deeper root cause all three share: `read_col_i64`'s silent default arm (internal.h:~164). Making it assert/reject unsupported types would have turned every one of these into a loud failure years ago — and the codebase now has four hand-rolled copies of the per-cell GUID hash/eq (this PR's, hs_hash_row/hs_eq_rows in collection.c, wide_key_hash_at/wide_key_eq_at in group.c), so a shared cell hash/eq helper would be the natural home for the fix.

Minor, test only: `guid_key.rfl` rebuilds the joins on every assertion (inner-join three times, full-join twice) — the sibling `nested_columns.rfl` binds once with `(set IJ ...)` and asserts on the binding; same shape would help here. Also its comment references `BUGS20260823.md #2`, which doesn't exist in the repo.

Happy to approve once the asof/window entries are either fixed or nyi-guarded for GUID (F64 ideally too) — the equi-join half is solid as-is.

belowzeroff:
Thanks for the thorough pass — the asof / window / F64 gaps are all addressed now (the follow-up commit landed just after this review).

Rather than extend the STR-only check, all three entry points now go through a **whitelist** of the types `read_col_i64` can actually represent:

```c
static bool join_i64_eq_key_supported(int8_t t) {
    switch (t) {
    case RAY_BOOL: case RAY_U8: case RAY_I16: case RAY_I32:
    case RAY_I64:  case RAY_DATE: case RAY_TIME: case RAY_TIMESTAMP:
    case RAY_SYM:  return true;
    default:       return false;   /* STR, GUID, F32, F64, object-shaped → nyi */
    }
}
```

wired into all three sites you flagged:
- `window_join_impl` fused path (was the STR-only check at ~:15212),
- `window_join_impl` legacy fall-through (~:15786),
- `ray_asof_join_core` (~:15906).

So GUID, F64 (and F32/STR) equality keys now fail loud with a typed
`nyi: … equality key column types <A> and <B> are not supported` instead of
comparing the wrong bytes — for all of `asof-join`, `window-join`,
`window-join1`, and the 4-arg compat forms.

Regression coverage in `test/rfl/join/str_key_nyi.rfl` pins exactly your
repros: asof + window/window-join1 → `nyi` for **STR, GUID, and F64** keys
(both operand orders), while SYM keys stay correct and order-independent and
the base equi-join still accepts STR keys (no over-rejection).

Test nits also done: `guid_key.rfl` now binds each join once (`(set IJ …)`)
and asserts on the binding, and the stale `BUGS20260823.md #2` comment is gone.

The branch is rebased onto the latest `dev`.

On the deeper points — a shared per-cell hash/eq helper, and making
`read_col_i64`'s default arm reject instead of silently truncating — fully
agree those are the right long-term home; I'd like to do that as a separate
sweep so this PR stays a focused correctness fix. PTAL / re-review when you get
a chance.

singaraiona (approved):
Re-verified the follow-up (ASan/UBSan build, suite 3706/3706, all original repros re-run). The whitelist approach lands exactly right:

- All three entry points (fused window path, legacy fall-through, asof core) now route through `join_i64_eq_key_supported`, applied to both sides' parted-unwrapped types — my three repros now produce typed `nyi` errors instead of `[0Nl ...]`/corrupt aggregates, one-sided mismatches (I64 vs F64) and mixed multi-key cases reject too.
- The whitelist is precisely the set `read_col_i64` represents correctly (checked arm-by-arm, including the width-aware SYM dict-id handling) — nothing misreadable is admitted, nothing readable is rejected. Positive cases confirm no over-rejection: SYM/I64/TIMESTAMP asof and window joins all still produce correct results, including the 4-arg compat forms.
- The equi-join GUID fix survives the rebase (inner/left/full/anti verified), the error-cleanup paths mirror the pre-existing STR checks with no leaks under ASan, and the test nits are done.

Agreed on keeping the shared cell hash/eq helper and the `read_col_i64` default-arm rejection as a separate sweep — this PR is a clean, focused correctness fix as it stands. Approving and merging.

PR #429, opened 2026-08-25
fix(format): bound fmt_ensure growth; loud failure past 2GiB output

belowzeroff:
## What a user sees today

Any code path that formats more than ~1 GiB of text through the shared formatter **hangs the process forever at 100% CPU** — no error, no output, must be killed. A single accidental `(println t)` on an unfiltered hundred-million-row table is enough. Exposed consumers include `println`/`show`, IPC message rendering (`src/core/ipc.c:507`), timer messages, and CSV error context (`src/io/csv.c:3217`).

### Reproduction

```lisp
(set v (til 120000000))   ;; ~120M i64 -> ~1.09 GiB of formatted text
(println v)
;; UBSan: src/lang/format.c:68:17: signed integer overflow:
;;        1073741824 * 2 cannot be represented in type 'int'
;; ...then the process spins at ~99% CPU forever (verified 40s+, zero bytes out)
```

## Root cause

`fmt_ensure` doubles its `int32_t new_cap` while `new_cap < len + extra`. Once `cap` reaches `2^30` (1 GiB) one more doubling overflows `int32`; with wrapping, `new_cap` walks negative and then to `0`, where `0 * 2 == 0` forever — the loop condition can never become false. UB plus a guaranteed infinite loop.

Additionally, both `ray_alloc` results in this buffer's lifecycle (`fmt_init`, `fmt_ensure`) were dereferenced unchecked, so deep OOM would crash on `ray_data(NULL)` anyway.

## Fix

- Compute the requirement in `int64`.
- Requirement above `INT32_MAX` (or negative `extra`) → clear message + `abort()`. The `fmt_*` helpers are all void, so there is no caller to return `ray_error` to; a loud stop beats a silent hang, and SIGABRT lets the existing crash handler dump diagnostics.
- Clamp the doubling to `INT32_MAX`: requirements between 1 GiB and 2 GiB now complete in one right-sized allocation (the old code UB'd exactly there).
- Check `ray_alloc` results in `fmt_init`/`fmt_ensure`.

## Verification

On the UBSan build:

| case | before | after |
|---|---|---|
| `(println (til 120000000))` (~1.09 GiB text) | infinite hang after UB report | prints 1,088,888,892 bytes, exit 0, no sanitizer reports |
| `(println (til 260000000))` (>2 GiB text) | same hang | `error: format: formatted output exceeds the 2GiB buffer limit` + fatal SIGABRT diagnostics |

Full suite green: 3701/3702 passed (1 skipped), exit 0.

No unit regression is added deliberately: any repro needs >1 GiB of formatted text (~40 s wall, multi-GB RSS) — out of scale for the rfl test suite.


singaraiona (changes requested):
Reviewed with a verification pass (17 candidates from 7 angles; 4 survived). The int64-widened growth math itself is sound — the doubling loop, INT32_MAX clamp, and cap store all check out, and NULL-checking ray_alloc is a strict improvement. But the headline guarantee has a hole:

**1. The `extra < 0` guard is unreachable — the wild-memcpy path this PR claims to close is still open.** For negative `extra`, `need = len + extra < len <= cap`, so the fast-path `if (need <= (int64_t)b->cap) return;` on the line above always returns first. Concretely: formatting a string whose length is in [2^31, 2^32) — `(int32_t)strlen(s)` in `fmt_puts` (line ~112) or `(int32_t)ray_str_len(s)` at the `fmt_putn` call sites (~255, ~404) wraps negative → `fmt_ensure` silently returns via the fast path → `memcpy(b->buf + b->len, s, (size_t)slen)` sign-extends to ~2^64 bytes → out-of-bounds write with no diagnostic. The check must precede the fast-path return.

**2. `abort()` on 2GiB/OOM is remotely triggerable and kills the whole server.** The IPC server evals client expressions and even formats the raw client payload for query logging (`ipc.c:507/553/560`), so a single client query like `(println <100M-element vector>)` — or any grow that ray_alloc can't satisfy under memory pressure — takes down every connected client. The codebase already has the designed recoverable path (`RAY_OOM_OBJ`, and every `ray_fmt` caller checks `RAY_IS_ERR`), so a sticky error flag on `fmt_buf_t` surfacing as an error from `fmt_to_str` is structurally possible with no signature changes. Loud is right; process-fatal is more than loud.

**3. The int32 length casts defeat the 2GiB loud-failure guarantee for single large arguments.** A string ≥4GiB wraps the cast to a small positive count and silently emits a truncated prefix — no abort, no error; [2^31, 2^32) hits finding 1's wild memcpy. Widening the lengths to int64 up to the fmt_ensure boundary closes both.

**4. Minor:** the `RAY_IS_ERR(b->block)` halves of the new alloc checks are dead — `ray_alloc` returns NULL on every failure path, never an error object; the branch encodes a false contract.

Findings 1+3 together mean the PR currently changes behavior only for the multi-small-append route to 2GiB; the single-large-argument routes still crash or truncate silently. Happy to re-review once the guard moves above the fast path and the casts are widened; the abort-vs-error question is a design call I'd like your take on either way.

belowzeroff:
Thorough — all four addressed in `48cab30b`.

**2 (abort → recoverable):** agreed, process-fatal is wrong here since `ray_fmt` runs on IPC-server threads formatting client-supplied values, so it's remotely triggerable. `fmt_buf_t` now carries a sticky `err` flag: `fmt_ensure` sets it on the 2GiB ceiling or a failed `ray_alloc`, the append primitives (`fmt_putc`/`fmt_puts`/`fmt_printf`/`fmt_putn`) no-op once it's set, and `fmt_to_str` surfaces it as a `RAY_IS_ERR` (`format` error). No `abort()` remains, and callers already check `RAY_IS_ERR`.

**1 (unreachable `extra < 0` guard):** moved above the fast-path return — for a negative `extra`, `need = len + extra < len <= cap` returned first, leaving the bad length to the wild memcpy. It's now the first check in `fmt_ensure`.

**3 (int32 length casts):** widened `fmt_ensure`'s parameter and the `fmt_puts`/`fmt_putn` lengths (including the `ray_str_len` sites — str atoms, list children, table cells) to int64, so a `[2^31,2^32)` or ≥4GiB argument reaches the ceiling check intact instead of wrapping to a small positive count and emitting a truncated prefix.

**4 (dead `RAY_IS_ERR(block)`):** removed at both alloc sites — `ray_alloc` only ever returns NULL — along with the `<stdlib.h>` that was only there for `abort()`.

Full ASan+UBSan suite green; normal and moderately large formatting round-trips unchanged. I didn't add a suite test for the 2GiB boundary itself — triggering it needs a multi-GB allocation (peak buffer ~2GiB), which would bloat CI memory/time. Happy to add one behind a large-test/opt-in gate if you'd prefer.

Unrelated aside: the run surfaced a pre-existing UBSan hit at `src/table/sym.h:64` (`index -1` in `ray_read_sym`) during `rfl/query/legacy_strl_predicates` — it's in the query path, independent of this diff, so flagging it separately rather than folding it in here.

singaraiona (approved):
Re-verified 48cab30b on an ASan/UBSan build, including a probe harness against both this head and the parent commit. All four fixes are genuinely in — and the parent-commit probe confirmed the original diagnosis: `fmt_putn(&b, "x", -5)` on the old code hits ASan `negative-size-param` in the memcpy; on this head it no-ops, sets the sticky err, and `fmt_to_str` surfaces the error.

- **Guard placement**: now first (after the sticky-err check — correct ordering). With int64 lengths end-to-end no in-tree caller can even produce a negative `extra` anymore, so it's clean defense-in-depth, not contradiction.
- **Sticky err**: no `abort` remains; `err` initialized on both `fmt_init` paths; all four primitives no-op after err with no spin risk. I audited every `fmt_buf_t` exit path — both `fmt_to_str` entry points surface `RAY_IS_ERR`, every `ray_fmt` caller checks it, and the print wrappers degrade safely (`ray_str_ptr` returns NULL/0 for error objects → zero-byte fwrite). Empirically: the 2GiB ceiling and a 2^33-length `fmt_putn` both come back as recoverable errors with no allocation.
- **int64 widening**: all five length sites widened; the remaining int32 casts are post-guard narrowing or small snprintf returns, all fine; no overflow path near INT64_MAX.
- **Dead checks**: gone at both sites, `<stdlib.h>` dropped.

Suite 3708/3708. On the 2GiB CI test: agreed, skip it — the repo has no large-test gate convention (only `RFL_UPDATE`), and inventing one for this isn't worth it; your harness-level coverage is the right call. The `sym.h:64` UBSan note you spotted is indeed the pre-existing exec_filter atom-predicate issue flagged on #385's thread — tracked there, unrelated to this diff by construction.

Two optional polish items, non-blocking: `ray_fmt_print`/`ray_fmt_pp_print` could print the error message instead of silently printing nothing on a format failure (REPL is the sole caller), and `fmt_table`'s per-cell tmp buffers don't propagate `tmp.err` outward (OOM-only, display-only, cells already truncate at 255 bytes).

Approving and merging.

PR #430, opened 2026-08-25
fix(sort): one rule for where a null sorts, across every sort path

### LINKED ISSUE #417: Null placement in sorts is inconsistent: SYM vs STR disagree, and SYM flips with row count
## Summary

Where nulls land in an ascending sort depends on the column type **and** on which sort path runs (row count):

| | ~24K rows (parallel path) | 6 rows (small path) |
|---|---|---|
| SYM `asc:` | nulls **first** | nulls **last** |
| STR `asc:` | nulls **last** | — |

Root cause: `sort_cmp` answers `nf ? -1 : 1` for a null operand with a `nulls_first` default derived from `desc`, while `sort_indices_ex` partitions nulls out and rotates them to the end chosen by `!desc` — two different defaults, and the type-specific paths pick different ones.

## Notes

- Pre-existing — reproduces identically on builds before the #416 work; surfaced while investigating whether the sort.c packed-key/sortedness gates could admit nullable SYM/STR (they can't be relaxed until this is settled, because a payload-only sortedness scan ranks sym-0 smallest while the real sort places nulls by policy).
- After bfb5b380 (empty SYM/STR are canonical nulls), this is user-visible on any sorted query over a text column with empties: the empties' position depends on data size.
- Fix needs a decision first: pick ONE null-placement policy (e.g. nulls first on asc, last on desc — or the kdb-style nulls-smallest), then align `sort_cmp`, `sort_indices_ex`, and every type arm, with tests pinning the matrix at both small and parallel sizes.
- The reproduction matrix is documented (not asserted) in `test/rfl/null/sym_str_null_compare.rfl`.

ser-vasilich:
Fixes #417

## The problem

Where a null sorts was spelled out separately in **five** places, and they did not agree:

| place | default for ASC |
|---|---|
| `sort_cmp` | nulls last (`nf = desc`) |
| `sort_indices_ex` | nulls first (`nf = !desc`) |
| `ray_sort_op` (query builder) | nulls last |
| `fused_topk.c` | nulls last, hardcoded |
| `apply_sort_take` | `desc` passed straight through as the flag |

Which one answered depended on the **row count** — a small input takes the comparison path, a large one the radix path — and on the **column type**, since the SYM, STR and numeric arms are separate. So the same query over the same data changed its answer as the table grew, and `(asc v)` over a vector disagreed with `select {asc: v}` over that column. Both behaviours were asserted as correct in the test suite.

## The fix

A null is the smallest value everywhere: ascending puts nulls first, descending last. `sort_nulls_first` in `ops/internal.h` is the one place that says so, and the paths take their default from it.

That default is what the bare verb, the radix arms and the rotate step already did, so what changes is the query sort, the fused top-k and take. **A top-k over a null-bearing column now returns the nulls**, since they are the smallest rows — the same answer kdb gives for `3#asc`.

Also removed: the per-key defaults array the merge fallback allocated only to recompute the same number, and its OOM path.

## Tests

`test/rfl/null/sort_null_placement.rfl` asks the same question of every path that can answer it — comparison and radix paths, SYM/STR/numeric columns, bare verb and query sort, and filter+top-k which runs the fused comparator. It is red before the fix and green after; verified by rebuilding the pre-fix tree, not by inspection.

Existing expectations that encoded the old answers are updated in the same commit as the fix, so **each commit is green on its own** and bisect stays honest.

## Known gaps, deliberately left

- The single source is a convention, not a guarantee: nothing in the type system stops a new path from spelling the rule out for itself, exactly as these five did.
- The SYM checks in the new test would pass under either policy, because an empty symbol is an in-band null that sorts smallest on payload alone. The case in #417's title — SYM flipping with row count — is documented rather than pinned; pinning it needs a SYM column carrying `HAS_NULLS`, which no literal or CSV read produces.
- The fused top-k still fills its per-key flag from the direction rather than from the caller's, since no caller can express one yet. An explicit `NULLS FIRST/LAST` clause in the query language would have to reach it, or it becomes the sixth place with its own opinion.
- `sort_cmp` is a hot comparison path and was not benchmarked; the change is logically no more expensive, but that is reasoning, not a measurement.


singaraiona (changes requested):
Reviewed with a verification pass — repros run empirically on both the PR head and a rebuilt base-dev binary, full suite 3706/3706. The core move (one `sort_nulls_first` rule, callers ask it) is right, and most paths genuinely converge on it. But one path escaped, and it turns the PR's own premise inside out:

**1. Blocker — the F64 multi-key direct-encode radix path contradicts the rule, and the PR makes it row-count-dependent where base was consistent.** The composite encoder (`radix_encode_fn`'s else-branch; F64 arm ~sort.c:1143-1148) reads raw bits with no null check, so NaN (the F64 null) encodes above all reals → nulls LAST on asc; `can_radix` has no HAS_NULLS gate and the multi-key `fits<=64` path never consults `nulls_first` at all. Measured with `select {from: T asc: [f g]}`, F64 key holding `0Nf` plus an I64 {0,1} second key (63+1 bits → direct-encode):

| rows | path | asc nulls |
|---|---|---|
| 60 | insertion/sort_cmp | **first** |
| 100 | encode+introsort | **last** |
| 5000 | encode+msd radix | **last** |

Base dev returned nulls-last at every size — internally consistent. The PR changed only the ≤64-row comparator default, so the same query now flips placement as the table grows, and F64 diverges from I64 multi-key (whose INT64_MIN sentinel happens to coincide with the new rule — spot-checked clean at all sizes). Notably the rank-then-compose fallback (total_bits>64) *is* null-aware — only the direct-encode arm is wrong. Fix: null-aware composite encode for F64 honoring `nulls_first`, or gate direct-encode on HAS_NULLS for F64 keys; either way please add F64 multi-key tests straddling the 64-row and radix thresholds.

**2. Must-fix in this PR — five comments still teach the OLD rule**, and stale policy comments were the mechanism of the original #417: `sort_indices_ex`'s header doc (~sort.c:2282, "nulls last for asc" — the central entry point's contract now states the opposite of what the code does), the fused-topk planner gate (~query.c:5881), `idxop.c:2071` + the `idx_fresh_nonull` rationale (~idxop.c:1201, now inverted — also note a sort index persisted pre-PR carries the old order; the null-free gate keeps that safe today but the comments are what the person lifting the gate will read), `internal.h:904` ("for single-key F64" — the field drives every typed arm), and the `group.c:2748` parity claim (see next point).

**3. Should-fix — grouped top/bot null policy is now mis-documented.** Grouped kernels skip nulls (`med_is_null` / `ray_valid_at`) while standalone `topk_take_vec` now includes them: `V=[3 0N 1 0N 2]` gives standalone `(bot V 2)` → `[0Nl 0Nl]` vs grouped → `[1 2]`. The divergence class predates this PR (on base, `top` diverged instead — the PR flipped which clause disagrees), and null-skipping is defensible aggregate semantics — but the comments claiming parity with `topk_take_vec` are now false on every clause. State the policy explicitly and fix the comments here; the behavior decision itself can be a follow-up. (Adjacent pre-existing oddity spotted while testing, identical on base: grouped `top v 4` with only 3 non-nulls returns them unsorted — the `kept < K` leg never sorts.)

**4. Minor — DESC back-fill coverage gap:** the rewritten `test_topk_null_sort_key_desc` asserts only non-null results, so no test puts a null INTO a DESC top-k output via the bounded-heap branch (non-null-count < k < len). Runtime is correct today (verified `top V 4` → `[3 2 1 0Nl]`, fused likewise) — one assertion would pin it.

Refuted for the record: the `descs` NULL-deref concern (docs never advertise NULL descs; pre-existing memcpy already dereferenced), the `fpk_keyspec_t.nulls_first` field (normal caching with a clear comment), and the removal-narrative comment complaints. The eagerly-materialized default `nfs[]` arrays and exec_sort's dead ternaries are optional cleanup.

Happy to re-review once the F64 encode path and the comments are fixed — with those done this genuinely delivers the one-rule promise.

singaraiona (approved):
Re-verified the follow-up commits on an ASan/UBSan build (suite 3712/3712, original repro re-run at every size). All four items are addressed:

- **Blocker — fixed.** The repro now yields nulls FIRST on asc (LAST on desc) at 60/100/5000 rows alike; the gate forces nullable-F64 keys onto the null-aware rank-then-compose fallback. I probed the attr-only concern hard — computed NaN (0/0, sqrt(-1), log(-1)), CSV empty-field F64, sliced and gathered columns — and every route stamps HAS_NULLS before the sort, matching the invariant the whole file already relies on (the single-key F64 arm gates on the same attr, and parted flatten ORs segment attrs in). F32 can't reach the packed encode; integer and non-null F64 keys still pack, so no perf loss. The size-straddling tests (comparator/packed/radix legs, both signs, integer twin) are exactly what I asked for.
- **Comments — all six fixed** (the five flagged plus ops.h:881), and I cross-checked the new text against the code: no new inaccuracies; group.c now honestly documents the grouped-skip-nulls policy instead of the false parity claim.
- **top/bot — the heap leg now agrees with the sort it falls back to**, with differential pins that fail on the old code. Leaving the grouped kernels on skip-nulls as documented policy is fine by me — the comment now tells the truth, which was the ask.
- **DESC back-fill — tested** (`test_topk_null_desc_backfill`, first coverage of that heap leg).

One discovery from the verification worth a follow-up issue (pre-existing, NOT introduced here — your gate actually removed the easiest trigger): the packing prescan hits UBSan on wide-span non-null F64 multi-keys — `kmax - kmin` signed overflow at sort.c:2702 and a `shift exponent 64` at :2704 (the `(1<<bits) <= range && bits < 64` loop evaluates the shift before the bound check). Results stay correct (fallback taken), but both expressions are UB; swapping the `bits < 64` test to the front and doing the range math unsigned would close it.

Approving and merging — this now delivers the one-rule promise for real.

Issue #411, opened 2026-08-16
update ... by: loses HAS_NULLS on the per-group gather — aggregates read null sentinels as values


singaraiona:
## Summary

`ray_update`'s `by:` branch gathers each group's rows into a per-group sub-table with `ray_vec_new` + raw `memcpy`, but never propagates `RAY_ATTR_HAS_NULLS` from the source column. The per-group aggregate then sees raw sentinel payloads as values.

## Repro

```lisp
(set NN (table [k v] (list [1 2 1 2] (as 'I64 [10 0N 30 40]))))
(at (update {mn: (min v) by: k from: NN}) 'mn)  ;; [10 0Nl 10 0Nl]  — wrong, should be 10, 40
(at (update {sm: (sum v) by: k from: NN}) 'sm)  ;; [40 -9223372036854775768 ...] — sentinel summed in
(select {mn: (min v) from: NN by: k})            ;; 10, 40 — correct oracle
```

## Notes

- Pre-existing on the atom-aggregate path (reachable on base via new columns); it became reachable on *existing* columns once #406 removed the duplicate-column bug, which is how it surfaced during that PR's review.
- Likely one-line fix adjacent to the gather in `ray_update`'s `by:` branch: `sub_col->attrs |= full_col->attrs & RAY_ATTR_HAS_NULLS;` — plus a regression test mirroring the repro (min/sum over a nullable column under `update by:`, checked against the `select by:` oracle).

singaraiona:
Confirmed on current `dev` at `9e1592a4` (v2.5.14 debug build).

The report's diagnosis is correct: the `update by:` per-group gather allocates each sub-column and raw-copies its payload, but does not preserve `RAY_ATTR_HAS_NULLS`. The copied I64 null sentinel is consequently treated as an ordinary value by `min`/`sum`.

I am implementing the fix with regression coverage for nullable grouped `min` and `sum`, checked against the corresponding `select by:` result.

singaraiona:
Implemented and verified locally.

The update-by grouped gather now preserves `RAY_ATTR_HAS_NULLS` whenever any selected source row is null, so aggregate kernels no longer take the null-free fast path on gathered data. Added `test/rfl/regress/issue_411.rfl` covering grouped `min` and `sum` plus select/update agreement.

[...]

PR #431, opened 2026-08-25
refactor(ops): drop unreachable ISNULL paths; pin load-select mask regression

protocolstardust:
## What this PR is now

Reworked per review onto current dev (a93763c3). The original uninitialized-ISNULL-mask bug this branch fixed was fixed more completely on dev by bfb5b380 (`exec_elementwise_unary` now answers `OP_ISNULL` for every input type before the typed dispatch). What dev still carried was the dead code that early-out orphaned, and no end-to-end repro of the original symptom. This PR supplies both:

- **Delete the unreachable I64→BOOL ISNULL kernel** in the typed dispatch — the `OP_ISNULL` early-out returns before any dispatch branch runs, so this kernel could never execute and merely duplicated a zero-fill.
- **Delete the unreachable ISNULL arm of the null-propagation pass** — same reason; the pass now unconditionally does `propagate_nulls` for the ops that reach it.
- **Add `test/rfl/regress/load_select_isnull_mask.rfl`** pinning the original load→select symptom: a 1000-row splayed table with a file `.sym` domain, a `load` that interns 60 fresh symbols (flipping the FILE-domain SYM column off the fused path into the typed fallback), then `nil?` / `not nil?` WHERE counts over SYM, I64 and F64 columns, cross-checked against per-element `map nil?`. Pre-bfb5b380 this returned 0/539/570/1000 of 1000 rows depending on allocator garbage; neither the original fix here nor bfb5b380 pinned the repro.

No vec.c change: the dead `case RAY_STR:` flagged earlier is already gone on dev — bfb5b380 made the SYM/STR payload-null checks live at the top of `ray_vec_is_null`.

## Validation

- Full suite on the rebased branch (UBSan build): 3710 of 3711 passed, 1 skipped, 0 failed.
- The new regression test fails meaningfully if the early-out regresses (verified it exercises the FILE-domain fallback trigger from the original report).

### DIFF
--- a/src/ops/expr.c
+++ b/src/ops/expr.c
@@ -2699,17 +2699,7 @@ ray_t* exec_elementwise_unary(ray_graph_t* g, ray_op_t* op, ray_t* input) {
             out_off += n;
         }
     } else if (in_type == RAY_I64 && out_type == RAY_BOOL) {
-        if (opc == OP_ISNULL) {
-            /* ISNULL over a non-null vec: always false here; the
-             * null-propagation pass at the end of the function sets
-             * dst[i]=1 for null rows of the input. */
-            while (ray_morsel_next(&m)) {
-                int64_t n = m.morsel_len;
-                uint8_t* dst = (uint8_t*)((char*)ray_data(result) + out_off);
-                for (int64_t i = 0; i < n; i++) dst[i] = 0;
-                out_off += n;
-            }
-        } else if (opc == OP_CAST) {
+        if (opc == OP_CAST) {
             /* (as 'BOOL i64_col) — truthy semantics; NULL_I64 = INT64_MIN
              * sentinel is non-zero but logically missing, so skip it. */
             while (ray_morsel_next(&m)) {
@@ -2854,18 +2844,10 @@ ray_t* exec_elementwise_unary(ray_graph_t* g, ray_op_t* op, ray_t* input) {
         }
     }
 
-    /* Propagate null bitmap from input to result.
-     * ISNULL is special: set output to 1 for null elements. */
-    if (vec_may_have_nulls(input)) {
-        if (op->opcode == OP_ISNULL) {
-            for (int64_t i = 0; i < len; i++) {
-                if (ray_vec_is_null(input, i))
-                    ((uint8_t*)ray_data(result))[i] = 1;
-            }
-        } else {
-            propagate_nulls(input, result, len);
-        }
-    }
+    /* Propagate null bitmap from input to result.  OP_ISNULL never reaches
+     * here — it is handled and returned before the typed dispatch above. */
+    if (vec_may_have_nulls(input))
+        propagate_nulls(input, result, len);
 
     /* OP_NEG/OP_ABS over i64: |INT64_MIN| and -INT64_MIN don't fit — surface
      * as typed null (by convention).  Loop above used unsigned wrap, so
--- a/test/rfl/regress/load_select_isnull_mask.rfl
+++ b/test/rfl/regress/load_select_isnull_mask.rfl
@@ -1,0 +1,51 @@
+;; Regression: load → select ISNULL WHERE-mask corruption.
+;;
+;; exec_elementwise_unary once had dedicated ISNULL kernels for only some
+;; input types; every other type fell through the typed morsel dispatch
+;; without writing a byte, so the BOOL result buffer came back as whatever
+;; the allocator recycled.  Trigger: a splayed table with a file .sym
+;; domain, queried after a `load` interned new symbols — the FILE-domain
+;; SYM column then bails from the fused expression path
+;; (EXPR_BAIL_SYM_DOMAIN) into the typed fallback, and
+;; `(select {where: (nil? sym)})` over 1000 rows returned 0, 539, 570 or
+;; 1000 rows depending only on the size of the previously loaded file.
+;;
+;; OP_ISNULL is now answered for every input type before the typed
+;; dispatch (expr.c exec_elementwise_unary); this file pins the original
+;; end-to-end symptom.
+
+(.sys.exec "rm -rf /tmp/rfl_isnull_mask") -- 0
+(.sys.exec "mkdir -p /tmp/rfl_isnull_mask") -- 0
+
+;; 1000-row table: SYM column cycling aa..dd (never null), I64 column with
+;; nulls at rows 0 and 999, F64 column with a null at row 500.
+(set SymM (at [aa bb cc dd] (% (til 1000) 4)))
+(set QtyM (concat (concat [0N] (+ 1 (til 998))) [0N]))
+(set PxM (as 'F64 (concat (concat (til 500) [0N]) (+ 501 (til 499)))))
+(set T-Mask (table [sym qty px] (list SymM QtyM PxM)))
+(.db.splayed.set "/tmp/rfl_isnull_mask/t" T-Mask "/tmp/rfl_isnull_mask/sym")
+
+;; Pad script: 60 `(set padN 1)` lines.  Loading it interns 60 symbols the
+;; file .sym domain doesn't have, reproducing the original trigger.
+(.sys.exec "awk 'BEGIN{for(i=1;i<=60;i++) printf \"(set pad%d 1)\\n\", i}' > /tmp/rfl_isnull_mask/pad.rfl") -- 0
+(load "/tmp/rfl_isnull_mask/pad.rfl")
+pad60 -- 1
+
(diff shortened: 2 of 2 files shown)


singaraiona (approved):

singaraiona (changes requested):
Reviewed with verification against the head (including confirming the fix mechanism is real: `ray_vec_new` does not zero the payload, so the uninitialized-mask bug exists and the memset closes it). Three findings, one of them a correctness gap the new comment actively papers over:

**1. SYM/STR payload-nulls still report non-null — and the new comment asserts the opposite.** `ray_vec_is_null` (vec.c ~1465-1471) treats SYM id 0 and zero-length STR as null "even when HAS_NULLS is absent", but the ISNULL fill pass (~expr.c:2767) is gated on `vec_may_have_nulls`, which checks only the attr. So for a persisted/gathered SYM column with null syms (or a STR column with empty strings) lacking `RAY_ATTR_HAS_NULLS`, the memset turns prior garbage into a deterministic all-false — better than garbage, but `(nil? col)` now confidently disagrees with every other consumer (format, morsel skip, agg) for those rows. The added comment claiming SYM/STR "correctly stay all-false, matching ray_vec_is_null" is factually wrong; the fill loop should run unconditionally for SYM/STR (or `vec_may_have_nulls` should consult payload semantics for those types).

**2. The I64→BOOL ISNULL kernel (~expr.c:2610) is now dead code** — it re-zeroes a buffer the new memset already zeroed. Keep one zero-fill (delete the kernel branch, or skip the memset for I64) rather than two paths that can drift.

**3. The comment claiming "Only I64 and F64 inputs have dedicated kernels below" is wrong for F64** — there is no F64 ISNULL kernel in the typed dispatch; F64 relies entirely on the new memset plus the propagation pass. A maintainer trusting the comment and removing the memset for F64 would reintroduce the exact corruption this PR fixes.

The memset itself is the right fix and can ship — but please fix the SYM/STR fill gap (or at minimum correct the comment so it doesn't certify the wrong behavior), and pick one zero-fill. Happy to re-review.

protocolstardust:
Addressed in 296b1e75 — thanks for the careful pass.

**#2 (dead I64 ISNULL kernel): fixed.** Deleted the kernel branch; the pre-zero memset is now the single zero-fill and the null-propagation pass the single place that sets 1s, so I64 follows the same path as every other type.

**#3 (comment claims an F64 kernel): fixed.** Confirmed there is no F64 ISNULL kernel in the typed dispatch (only `F64→F64`, `F64→I64`, and CAST branches exist). With the I64 kernel gone, the comment now states that no dispatch branch writes ISNULL output — the memset plus the propagation pass are the whole story, so there is nothing left to mistakenly "optimize away" per-type.

**#1 (SYM/STR payload-nulls): respectfully disagree — I believe this one is a misread of the current head.** `ray_vec_is_null` returns false for SYM/STR unconditionally at the top of the function (src/vec/vec.c:1441, "SYM and STR columns are no-null by design"), before the sentinel switch is ever reached — the `case RAY_STR:` inside that switch is unreachable leftover from before 544d1767 ("STR has no null") and 564996c3 ("remove dead SYM null branches"), and is probably what suggested payload-null semantics. I also checked format.c, agg.c, and morsel.c: none of them special-case sym id 0 / empty STR as null. So all-false from `(nil? col)` on a persisted SYM/STR column *is* consistent with every other consumer, and the comment's claim stands. Happy to remove the dead `RAY_STR` switch case in a separate cleanup if you'd like, but I kept it out of this PR.

Verified against the release build: splayed sym col `(nil? s)` WHERE → 0 rows, I64 col with two 0N → 2 rows, F64 col with one 0Nf → 1 row; full suite 3631/3632 pass (1 skipped, 0 failed).

singaraiona:
Dug into the disagreement with builds of both trees, and the resolution is interesting: **we were both right — about different versions.** Your citation is accurate for this branch's base (dev @ 9961388c, July 16): the early return `if (type == RAY_SYM || RAY_STR) return false;` is there, the `case RAY_STR:` below it is dead leftover from 544d1767/564996c3 exactly as you said, and on this tree `nil?` over SYM/STR being all-false is internally consistent — I reproduced your results on your head.

But the branch was never rebased onto current dev (GitHub reports it CONFLICTING), and dev moved underneath it: **bfb5b380 "fix(core): normalize null and schema handling" (Aug 17, fixes #411/#414/#415) reversed the SYM/STR policy** — it deleted that early return and added exactly the code my review quoted ("SYM/STR nulls are canonical empty payloads. Check them directly even when HAS_NULLS is absent", SYM id 0 / zero-length STR). Same probe script on dev tip: `nil? s` → `[false true false false]`, the empty sym prints as `null`, and `where (nil? s)` selects the row. So the divergence I flagged is real on the merge target — my review was reading dev, yours was reading the base. Classic version split.

The bigger consequence: bfb5b380 also already fixed this PR's headline bug, more completely — dev's `exec_elementwise_unary` has an OP_ISNULL early-out before the typed dispatch that consults `ray_vec_is_null` for SYM/STR/nullable inputs and memsets the rest. After the mandatory rebase, this PR's memset and comment become dead code for ISNULL (dev returns before reaching them), and the comment's SYM/STR sentence becomes false.

What's still genuinely useful from your work here: dev still carries the now-unreachable I64→BOOL ISNULL kernel (~expr.c:2702 on dev) and the unreachable ISNULL propagation arm (~2860) — your 296b1e75 deletion is exactly the right cleanup, just aimed at the wrong base. My suggestion: rework this PR into a small dead-code cleanup against current dev (delete the unreachable kernel + propagation arm, plus the dead `case RAY_STR:` in vec.c you offered to remove), or close it as superseded by bfb5b380 — same shape as #419/#409. If you rework it, a regression test for the original garbage-WHERE-mask symptom would be a nice addition since neither this PR nor bfb5b380 pinned that exact repro.

Sorry for the whiplash — the review should have flagged the base skew explicitly. Findings 2 and 3 were cleanly fixed in 296b1e75 on your tree; they just land in a different place after the rebase.

protocolstardust:
Reworked as suggested — thanks for running both trees down; the base-skew explanation resolved it cleanly.

The branch is now rebased onto current dev (a93763c3) with a single commit, 73db07a6:

- Dropped the unreachable I64→BOOL ISNULL kernel and the unreachable ISNULL propagation arm in `exec_elementwise_unary` — the `OP_ISNULL` early-out from bfb5b380 is now the single ISNULL path.
- The dead `case RAY_STR:` in `ray_vec_is_null` turned out to already be gone on dev (bfb5b380 made the SYM/STR payload checks live at the top of the function), so no vec.c change.
- Added `test/rfl/regress/load_select_isnull_mask.rfl` pinning the original end-to-end symptom: splayed table with file `.sym` domain, `load` interning 60 fresh symbols to force the FILE-domain column off the fused path, then `nil?`/`not nil?` WHERE counts over SYM/I64/F64, cross-checked against per-element `map nil?`.

PR title/body updated to match. Full suite on the rebased branch: 3710 of 3711 passed (1 skipped, 0 failed). Ready for re-review.

singaraiona (approved):
Re-verified the rework (ASan/UBSan build, full suite green). This is exactly the dead-code cleanup we landed on, and it's clean:

- **Single commit, clean rebase**: merge-base with dev is the current dev head (a93763c3), `dev..pr431` is exactly one commit and `pr431..dev` is empty — a clean fast-forward, and GitHub now reports mergeable (the earlier CONFLICTING is gone).
- **The removed paths are genuinely unreachable and their removal is behavior-preserving**: bfb5b380's `OP_ISNULL` early-out returns before the typed morsel dispatch, so both the I64→BOOL kernel and the tail propagation arm (both ISNULL-gated) were dead. Confirmed no non-ISNULL opcode depended on either — the I64→BOOL else-if now only serves OP_CAST and the chain stays well-formed. The original memset is gone too.
- **vec.c correctly untouched** — the dead `case RAY_STR:` was already removed by bfb5b380 on dev.
- **The regression test has teeth** — verified empirically: neutralizing the early-out and rebuilding makes `load_select_isnull_mask` fail with the exact documented symptom (`(count (select {where: (not (nil? sym))}))` → 539 instead of 1000). And I spot-checked `nil?` end-to-end (both `map nil?` and select-where) across I64/F64/F32/I32/I16/DATE/TIME/TIMESTAMP/SYM/STR/BOOL/GUID — all correct, no type lost coverage.

Full sanitizer suite: 3714/3714, 0 failed. (Minor: the PR body's "3710/3711, 1 skip" is stale vs the current dev tree — cosmetic, no action needed.)

Nice turnaround on the version-split — thanks for taking the rework. Approving and merging.

PR #432, opened 2026-08-25
fix(builtins): reject out-of-range numeric string casts

belowzeroff:
## Summary
- reject clean string values outside i32/i16/u8 target ranges instead of silently wrapping them
- keep existing non-string narrow cast wrap semantics intact
- add regression coverage for atom and list string casts

## Verification
- reproduced on origin/dev: (as 'u8 "256") -> 0x00, (as 'u8 "-1") -> 0xff, (as 'i16 "40000") -> -25536, (as 'i32 "3000000000") -> -1294967296
- make
- make test: 3701 of 3702 passed (1 skipped, 0 failed)


singaraiona (changes requested):
Reviewed with an empirical verification pass on the built branch. The range checks are right in shape — `strtoll` + explicit bounds is the correct mechanism, coverage of the cast targets is complete, and vector casts route through the same code. One confirmed gap, exactly at the boundary the PR exists to police:

**The checks admit INT32_MIN/INT16_MIN, so a clean in-range boundary string silently becomes a null atom.** Verified on the branch:

- `(as 'i32 "-2147483648")` → `0Ni`
- `(as 'i16 "-32768")` → `0Nh`
- `(as 'I16 (list "-32768" "1"))` → `[-32768 1]` — the vector path keeps the real value

The strict `<` at the two check sites lets the exact minimum through, but that bit pattern *is* the atom null sentinel (`ray_atom_is_null_fn`), so the same string yields a null scalar and a real vector element. That's precisely the "silent surprise on string ingest" class this PR closes elsewhere, and neither boundary is pinned by a test. Two defensible fixes: reject the sentinel minimum for atom casts with the same range error (my preference — the atom genuinely cannot represent it), or keep the aliasing and pin it with tests plus a comment so it's a documented quirk rather than an accident. Either way the vector/atom divergence deserves a test.

For the record, refuted along the way: the missing `!sp` guards are harmless (`ray_str_ptr` never returns NULL, even for empty strings), and `strtol`→`strtoll` loses nothing on any platform. Suite is green apart from the pre-existing `types/version_getters` skew unrelated to this diff. Happy to approve once the boundary is decided and pinned.

belowzeroff:
Good catch — verified and fixed in `92ab63be`.

The scalar string cast now rejects the exact sentinel minimum (`INT16_MIN` / `INT32_MIN`, and the same latent case for `INT64_MIN`) with a `domain` error instead of silently yielding `0Nh` / `0Ni` / `0Nl`.

On the atom/vector divergence: rather than keep and document it, I eliminated it. `cast_vec_numeric` parses each element through the same scalar `ray_cast_fn`, so `(as 'I16 (list "-32768" "1"))` now rejects too — consistent with this PR's "string ingest is loud" rule, while a numeric narrowing cast still maps a produced sentinel to null via `cast_mark_output_sentinels` (the documented narrow-int rule). So there's no longer an atom/vector divergence to pin.

`type/as.rfl`: flipped the two boundary cases that had pinned the old `0Nl` aliasing, and added the i64 sentinel, the MIN+1/MAX boundaries, and the vector rejection. Full ASan+UBSan suite green.

singaraiona (approved):
Re-verified the follow-up on an ASan/UBSan build (suite 3708/3708, all boundaries probed live). This landed better than the review asked:

- All three sentinel minimums now reject with a consistent `domain` error — and the i64 case closes a real pre-existing hole I hadn't flagged: `strtoll("-9223372036854775808")` returns INT64_MIN with no ERANGE, so it previously produced `0Nl` silently. Check ordering (ERANGE first, then the exact-MIN check) is right.
- The atom/vector divergence is resolved by making both loud — vectors route through the same scalar parser, and I confirmed the policy is the defensible one: an I16 vector holding a "real" −32768 is in-band ambiguous anyway (`at` collapses it to `0Nh` while `sum` treats it as the value), so rejecting it at string ingest is coherent, and it's explicitly tested.
- MIN+1/MAX parse to real values everywhere; u8/whitespace/hex/empty-string behaviors are untouched; numeric (non-string) narrowing semantics unchanged.

Two optional follow-ups, non-blocking: an I64-vector sentinel rejection test (behavior verified, just not pinned), and CSV typed ingest still accepts −32768 into an I16 column (`csv.c` boundary check) — pre-existing, but now the one remaining quiet path for the sentinel; worth aligning with this PR's loud rule in a future sweep.

Approving and merging.

PR #433, opened 2026-08-25
fix(arith): exact int64 floor division in scalar div

belowzeroff:
## Summary

Scalar/atomic `div` (`ray_idiv_fn`) routed the quotient through `double`, which:

- **lost precision above 2^53** — the identity `(div x 1) == x` broke for any
  `|x| > 2^53` (e.g. `(div 9007199254740993 1)` → `9007199254740992`);
- **hit signed-conversion UB** at the INT64 extremes: `(double)INT64_MAX`
  rounds up to exactly `2^63`, so the old `q > (double)INT64_MAX` guard never
  fired and `(int64_t)q` was evaluated on `q == 2^63`.

## What it looks like from the user's side

Nothing errors — the result is just silently wrong, and only for large
integers, so it slips through casual testing:

- Dividing a large 64-bit **id / hash / nanosecond timestamp** and getting a
  value that is off by 1–8. An id like `9007199254740993` comes back as
  `9007199254740992`; a later lookup / join / `in` test on that id then fails
  to match, or matches the wrong row.
- `(div a 1)` — a no-op the user expects to be the identity — **changes the
  number** for any `|a| > 9007199254740992`.
- At the 64-bit boundary the result is either a garbage number or a surprise
  null: `(div -9223372036854775807 -1)` → garbage (plus an
  `arith.c:286` sanitizer report), `(div 9223372036854775807 -1)` → `0Nl`.
- Any integer-heavy pipeline working with magnitudes beyond ~9×10¹⁵
  (scientific counts, fixed-point money in the smallest unit, packed keys)
  can silently corrupt a column with no diagnostic.

| input | before | after |
|---|---|---|
| `(div 9007199254740993 1)` | `9007199254740992` | `9007199254740993` |
| `(div 123456789012345678 1)` | `123456789012345680` | `123456789012345678` |
| `(div -9223372036854775807 -1)` | UBSan + garbage | `9223372036854775807` |
| `(div 9223372036854775807 -1)` | `0Nl` | `-9223372036854775807` |

## Fix

`ray_idiv_fn` now splits on operand kind:

- **Both integer** → divide in int64 space with a floor-toward-−∞ correction
  (`if ((av ^ bv) < 0 && q * bv != av) q--;`), mirroring the temporal path in
  `ray_mod_fn`. Exact for all magnitudes. `INT64_MIN / -1` (the only i64
  division overflow) is guarded to null — unreachable after the null check
  since `INT64_MIN` is the i64 null sentinel, but kept defensively.
- **Float operand(s)** → keep the `floor(as_f64(a)/bv)` path (preserves the
  existing float-div semantics), with the corrected guard `q >= 2^63`.

## Scope (what this PR does *not* cover)

This fixes the **scalar/atomic** path (`ray_idiv_fn`) — a bare `(div a b)` and
`RAY_FN_ATOMIC` mapping over vectors. `div` used **inside a `select`/query**
compiles to a *separate* fused kernel (`OP_IDIV` in `src/ops/expr.c`), which
still evaluates through `double` and therefore keeps the same precision loss —
e.g. `(select {x: (div bigcol 1) from: t})` over an i64 column `> 2^53` is
still wrong. That fused path is a distinct, still-open issue and is filed as a
follow-up; it is intentionally out of scope here to keep this change minimal
and isolated to `arith.c`.

## Tests

All existing `div` goldens in `test/rfl/integration/math.rfl` are unchanged
(integer floor semantics, mixed int/float, vector operands, null / divide-by-
zero → `0Nl`, temporal → type error). New goldens pin the `>2^53` and
INT64-extreme cases. Full ASan+UBSan test suite green (3648/3649, 0 failed) on
the `dev` base.


belowzeroff:
Superseded by #403 (`66a80fbe`), already merged to `dev`: it lands the identical exact-int64 `ray_idiv_fn` fix (same >2^53 and INT64-extreme behaviour, same floor-toward--inf correction), plus the compiled/fused `OP_IDIV` path and the precision goldens in `math.rfl`. This branch was cut from an earlier `dev`, so rebasing would add nothing and would revert the newer scalar overflow-UB fix (#388, `caa529d2`). Closing rather than rebasing.

PR #434, opened 2026-08-25
fix(storage): reject unsafe splayed column names

belowzeroff:
## Summary

- Reject unsafe splayed column names during save validation instead of silently skipping them.
- Keep the save loop fail-closed if an unsafe name reaches it.
- Replace the old skip-behavior tests with regression coverage for leading dot, slash, backslash, and embedded NUL names.

## User impact

Before this fix, a user could successfully run `.db.splayed.set` on a table that contained a column whose persisted filename was unsafe, such as `.hidden`, `a/b`, `a\b`, or a name containing an embedded NUL. The command returned success, but the unsafe column was not written to disk and was omitted from `.d`, so a later `.db.splayed.get` loaded a table with fewer columns than the user saved.

From the user's point of view this looked like silent data loss rather than an invalid table schema.

After this fix, the same save fails with a domain error before creating or updating the splayed directory, so successful saves preserve every user column.

## Tests

- `git diff --check origin/dev..HEAD`
- `make test DEBUG_CFLAGS='-fPIC -Wall -Wextra -Werror -Wstrict-prototypes -Wno-unused-parameter -std=c17 -g -O0 -march=native -DDEBUG' DEBUG_LDFLAGS=`
  - `3703 of 3704 passed (1 skipped, 0 failed)`


singaraiona (approved):
Reviewed with a verification pass — no correctness issues, approving. The ordering that the new test depends on holds: `splay_validate_persisted_names` runs before `ray_mkdir_p` and any symfile write, so a rejected save leaves no partial state on disk. All four callers propagate the `ray_err_t`, and this brings the materialized path in line with the streaming CSV→splayed writer, which already rejected unsafe names with the same error. Suite 3707/3707 at the head.

Two non-blocking cleanups for the same commit or a follow-up:

1. **The in-loop unsafe-name check (~splay.c:283) is now unreachable** — the up-front validation already returns for every unsafe name before the column-write loop runs, so the new cleanup/return block there can never execute (and can never be covered by a test). Either delete it in favor of the up-front check, or mark it explicitly as defense-in-depth; as written it reads as the live rejection path and misleads about where enforcement happens.

2. **Stale comment at ~splay.c:176**: the ".sym" collision guard's rationale still says dot-led names are "skipped below" — that skip is exactly what this PR removed. It should say rejected.

belowzeroff:
Thanks — both addressed in `793cd4e3`.

1. Removed the in-loop `splay_col_name_safe` check. `splay_validate_persisted_names()` runs up-front in the same `splay_save_impl` and there's no other entry point, so it was unreachable with no real defense-in-depth value — I deleted it rather than annotate it, so enforcement is single-sourced, and left a one-line note at the loop pointing to the up-front validation.
2. Fixed the `.sym` collision comment: dot-led names are "rejected up-front", not "skipped below".

No behavior change; the dot / slash / backslash / NUL rejection tests are unchanged and green.

Issue #78, opened 2025-03-03
Date type serialized as i32 over ipc


demyan:
Peer1
```
↪ (write h (list 'f 2025.03.03))
↪ (write h (list 'f 2020.03.03))
```
Peer2:
```
↪ (set f (fn[x] (println "type=% val=%" (type x) x)))
@f
↪ type=i32 val=9193
type=i32 val=7367
```

PR #435, opened 2026-08-25
fix(csv): reject impossible calendar dates in typed DATE/TIMESTAMP parse

belowzeroff:
## Summary

Typed CSV ingest (`.csv.read [DATE …]` / `[TIMESTAMP …]`) accepted **calendar-impossible dates** and silently rewrote them into a *different real day* instead of nulling them. This is bug **#78** from `bugs20260824.md`.

`fast_date()` in `src/io/csv.c` bounded the day only with `d <= 31`. A well-formed `YYYY-MM-DD` naming a day that cannot exist in its month therefore passed validation, and `civil_to_days()` then normalized it to a valid-but-wrong date.

A CSV with a typo'd date — `2024-02-31` (February 2024 has only 29 days):

```
$ cat trades.csv
sym,dlv
AAPL,2024-02-31
MSFT,2024-02-29
```

**Before** — the impossible date is silently "corrected" to a real, wrong day. No error, no null; the row now claims a delivery date two days later than anything the file contained:

```clojure
(at (.csv.read [SYM DATE] "trades.csv") 'dlv)
[2024.03.02 2024.02.29]      ;; 2024-02-31 became 2024.03.02 — silent data corruption
```

**After** — the impossible date reads as null (`0Nd`), exactly like an out-of-range month or day already did; valid dates (including the leap day) are unchanged:

```clojure
(at (.csv.read [SYM DATE] "trades.csv") 'dlv)
[0Nd 2024.02.29]
```

The same applied to `[TIMESTAMP]` columns, since `fast_timestamp()` reuses `fast_date()` for the leading date:

| Input (typed column) | Before | After |
|----------------------|--------|-------|
| `2024-02-31`             | `2024.03.02`                       | `0Nd` (null) |
| `2024-04-31`             | `2024.05.01`                       | `0Nd` (null) |
| `2023-02-29` (non-leap)  | `2023.03.01`                       | `0Nd` (null) |
| `2024-02-31T00:00:00`    | `2024.03.02D00:00:00.000000000`    | `0Np` (null) |
| `2024-02-29` (leap)      | `2024.02.29`                       | `2024.02.29` (unchanged) |

Why it matters: the corrupted value is indistinguishable from a legitimate date, so it survives into downstream joins, as-of lookups and aggregations — silently, with no error to alert the user.

## Root cause & fix

Validate the day against the actual, leap-aware length of the month. The month is still range-checked (`1..12`) **before** the days-in-month table lookup, so the index stays in bounds.

```c
if (RAY_UNLIKELY(m < 1 || m > 12 || d < 1)) { *is_null = true; return 0; }
static const int md[] = {0,31,28,31,30,31,30,31,31,30,31,30,31};
int leap = (y % 4 == 0 && (y % 100 != 0 || y % 400 == 0));
if (RAY_UNLIKELY(d > md[m] + (m == 2 && leap ? 1 : 0))) { *is_null = true; return 0; }
```

This mirrors the strict `(as 'DATE str)` cast hardened in #382. The CSV reader has its own `fast_date`/`fast_timestamp` and was missed both by that change and by #422 (which fixed separator/trailing-garbage strictness but not calendar validity).

## Tests

Extended `test/rfl/io/csv_types.rfl` (new section `18b`) with calendar-validity coverage for both `[DATE]` and `[TIMESTAMP]`, including leap-year edge cases:

- `2100-02-29` → null (century year, not divisible by 400 → not a leap year)
- `2000-02-29` → valid (divisible by 400 → leap year)
- 31st of a 30-day month (Apr/Jun) → null
- leap day `2024-02-29` and `2024-01-31` → parse unchanged

Full suite green under the default ASan/UBSan debug build: `3701 of 3702 passed (1 skipped, 0 failed)`.


singaraiona (approved):
Reviewed with a verification pass. The diff is correct and I'm approving — an exhaustive sweep (years 1–9999, all m/d 0–99) against a `timegm` reference found zero mismatches; the `m>12` guard fully protects the `md[]` index; every reject path sets `*is_null`; `fast_timestamp` inherits the validation; `detect_type` shares the same parser so inference stays consistent (impossible dates in sampled rows now infer STR, matching the #422 strict-grammar design); and the writer can only emit real dates, so round-trip is safe. The "mirrors ray_cast_fn()" claim checks out against builtins.c line-for-line.

Two follow-ups worth doing while this area is warm — neither blocks this PR:

**1. The language literal parser still has exactly this bug.** `src/lang/parse.c` (~line 263) keeps the coarse `day <= 31` check and `ymd_to_date` then normalizes: verified on the built binary, the literal `2024.02.31` evaluates to `2024.03.02`, `2023.02.29` → `2023.03.01`, and timestamp literals inherit it. That leaves impossible-date handling three-way inconsistent: CSV ingest → null, `(as 'DATE str)` → error, source literal → silently shifted day. A user typing `(where (= d 2023.02.29))` matches the wrong rows with no diagnostic — the same silent-corruption class this PR eliminates for CSV, one layer up.

**2. This is now the third verbatim copy of the month-length table + leap formula** (`dt_md` and `ts_md` in builtins.c:~1620/~1770 are identical), while `src/lang/cal.h` is the designated shared calendar header (`date_leap_year`, `MONTHDAYS`) that none of the three parsers use. A `days_in_month(y, m)` (or `ray_valid_ymd`) in cal.h, adopted by all three sites — and by the parse.c fix from point 1 — turns the next calendar tweak from a four-site hunt into one edit. Minor in the same vein: the two reject-ifs can merge into one (the `||` short-circuit already guards the index), which also drops the duplicated null-return body.

Merging this as-is; happy to review the parse.c + cal.h follow-up whenever.

PR #436, opened 2026-08-26
fix(store): let a mapped string column's pool own the region they share

### LINKED ISSUE #426: Engine bug — `select … where` over a splayed-loaded table leaves a `str` column with a dangling string pool
Found 2026-08-23 at
engine `v2.5.15` (`605723d6`), the pinned release; not re-tested on `dev`.

---

**Title:** fix(mem): `select … where` on a table loaded with `.db.splayed.get`
yields a `RAY_STR` column whose `str_pool` is released twice — SIGSEGV in
`ray_release_owned_refs` (`heap.c:896`) on rebind or at exit

## Summary

A table that carries a string (`str`) column is written with
`.db.splayed.set` and read back with `.db.splayed.get`. A `select` over the
loaded table with a `where` clause that keeps a **proper subset** of its
rows returns a table whose `str` column does not own its string pool: it
shares the loaded column's pool without retaining it. Two consequences,
both a SIGSEGV at `src/mem/heap.c:896` in `ray_release_owned_refs` (the
`v->type == RAY_STR && v->str_pool && …` line):

1. **Immediate** — if the loaded table is released first (the usual shape:
   the global holding it is rebound to the selection, then the selection is
   mutated), the next operation that rebinds/frees the selection reads the
   freed pool: `insert` into it crashes before it prints anything.
2. **At exit** — if both stay bound, every read of the selection works
   (`print`, `at`, `count`, element access all return correct data), and
   the process faults in `ray_env_destroy` when the environment releases
   both tables — the pool is freed twice.

Only `str` columns are affected. Tables with only `sym` and numeric
columns survive the same sequence; the same `select` over a never-splayed
in-memory table is fine; a `where` that keeps **every** row is fine; and
an `insert` into the loaded table *before* the `select` makes everything
fine (the insert copies the column into ordinary memory). The reproducer
below is deterministic — it crashes on every run.

## Environment

- rayforce `v2.5.15` (`605723d6b44d95d41244c9abd27fa99bae0d1ec6`), the
  `rayforce` binary built from the release tag (this build carries ASan;
  the fault is the engine's own SIGSEGV banner, not an ASan report).
- Linux x86_64 (Ubuntu, kernel 7.0.0-28-generic), glibc 2.39.
- Reproduced identically through the embedded runtime (`rayforce-rs`
  `c0a0bcc`) from a host process.

## Minimal reproducer

`repro.rfl` (adjust `dir`; it must be writable and end with `/`):

```lisp
(set 'dir "/tmp/rf-splayed-str/")
(set 't (table ['id 'note] (list [1 2 3] (list "one" "two" "three"))))
(.db.splayed.set dir t)
(set 'u (.db.splayed.get dir))

;; Case 1 — immediate: rebind the global to a proper-subset selection,
;; then mutate the selection. Nothing is printed; SIGSEGV.
(set 'u (select {from: u where: (not (== id 1))}))
(insert 'u (table ['id 'note] (list [9] (list "nine"))))
(print (at u 'note))
```

```text
$ rayforce repro.rfl

=== rayforce fatal SIGSEGV at fault addr 0x0000708bcda19062 ===
rayforce 2.5.15 (605723d6)
```

Case 2 — the same selection kept beside the loaded table: every read is
correct, the fault moves to exit.

```lisp
(set 'dir "/tmp/rf-splayed-str/")
(set 't (table ['id 'note] (list [1 2 3] (list "one" "two" "three"))))
(.db.splayed.set dir t)
(set 'u (.db.splayed.get dir))
(set 'v (select {from: u where: (not (== id 1))}))
(print (at v 'note))          ;; prints ["two" "three"] — correct
(print "end-of-script")       ;; printed
```

```text
$ rayforce repro2.rfl
["two" "three"]end-of-script
=== rayforce fatal SIGSEGV at fault addr 0x00007c4fa1a1b062 ===
rayforce 2.5.15 (605723d6)
$ echo $?
139
```

## Backtraces (gdb, `ASAN_OPTIONS=detect_leaks=0`)

Case 1 — the crash is the release of the OLD binding inside `insert`'s
`ray_env_set`: the selection being replaced owns a `str_pool` that was
freed when the loaded table went away on the first rebind.

```text
#0  ray_release_owned_refs (v=0x7ffff0009280) at src/mem/heap.c:896
896	    if (v->type == RAY_STR && v->str_pool && !RAY_IS_ERR(v->str_pool))
#1  ray_free (v=0x7ffff0009280) at src/mem/heap.c:1515
#2  ray_release (v=0x7ffff0009280) at src/mem/cow.c:72
#3  ray_release_owned_refs (v=0x7ffff0009200) at src/mem/heap.c:934
#4  ray_free (v=0x7ffff0009200) at src/mem/heap.c:1515
#5  ray_release (v=0x7ffff0009200) at src/mem/cow.c:72
#6  ray_release_owned_refs (v=0x7ffff0009100) at src/mem/heap.c:926
#7  ray_free (v=0x7ffff0009100) at src/mem/heap.c:1515
#8  ray_release (v=0x7ffff0009100) at src/mem/cow.c:72
#9  env_bind_global_impl (sym_id=402, val=0x7ffff0008d80, is_user=1) at src/lang/env.c:458
#10 env_bind_global_user (sym_id=402, val=0x7ffff0008d80) at src/lang/env.c:497
#11 ray_env_set (sym_id=402, val=0x7ffff0008d80) at src/lang/env.c:709
#12 ray_insert (args=0x7ffff0007ea8, n=2) at src/ops/query.c:14025
#13 ray_eval (obj=0x7ffff0007e80) at src/lang/eval.c:3929
#14 ray_do_fn (args=0x7ffff0005a28, n=7) at src/lang/eval.c:1922
#15 ray_eval (obj=0x7ffff0005a00) at src/lang/eval.c:3929
#16 ray_repl_run_file (…) 
#17 main (argc=2, argv=…) at src/app/main.c:250
```

Case 2 — the same line, reached from environment teardown:

```text
#0  ray_release_owned_refs (v=0x7ffff0008780) at src/mem/heap.c:896
#1  ray_free (v=0x7ffff0008780) at src/mem/heap.c:1515
#2  ray_release (v=0x7ffff0008780) at src/mem/cow.c:72
#3  ray_release_owned_refs (v=0x7ffff0008700) at src/mem/heap.c:934
#4  ray_free (v=0x7ffff0008700) at src/mem/heap.c:1515
#5  ray_release (v=0x7ffff0008700) at src/mem/cow.c:72
#6  ray_release_owned_refs (v=0x7ffff0008600) at src/mem/heap.c:926
#7  ray_free (v=0x7ffff0008600) at src/mem/heap.c:1515
#8  ray_release (v=0x7ffff0008600) at src/mem/cow.c:72
#9  ray_env_destroy () at src/lang/env.c:267
#10 ray_lang_destroy () at src/lang/eval.c:3576
#11 ray_runtime_destroy (rt=0x7ffff7ed1020) at src/core/runtime.c:414
#12 main (argc=2, argv=…) at src/app/main.c:310
```

(Under gdb the ASan build needs `ASAN_OPTIONS=detect_leaks=0`, or
LeakSanitizer refuses to run under ptrace and gdb reports "No stack".)

## What was varied, and what happened

Same script shape as case 2 unless stated; "fault" means SIGSEGV at exit
with the backtrace above, "ok" means exit 0 with correct output.

| variation | result |
|---|---|
| `select` with a `where` dropping the first / middle / last row | fault / fault / fault |
| `select` with a `where` that keeps every row (`(> id 0)`) | ok |
| `select` with no `where` | ok |
| the selection's `str` column never touched (only `(at v 'id)` read) | fault — the column's existence is enough |
| string lengths 3, 8, 12, 16, 20, 47 characters | fault at every length |
| table with `sym` + `long` columns only (no `str`) | ok |
| table with `long` columns only | ok |
| the same `select` over the in-memory table `t` (never splayed) | ok |
| `insert` a row into the **loaded** table `u` first, then `select` + `insert` | ok — the insert materialises the column |
| `insert` a row into the **selection** `v` while `u` stays bound | ok at exit too — the insert reallocates the selection's column, so nothing is shared any more |
| rebind the loaded table's global to the selection, then `insert` (case 1) | immediate fault, nothing printed |
| wider table: `long`, `timestamp`, three `sym`, `long`, `str` | same two outcomes as the two-column table |

The `.db.splayed.set` / `.db.splayed.get` calls are the two-argument
forms (directory only); the directory gets `.d`, one file per column and
a `.sym` beside them.

## Expected

The `str` column of a `select` result is an ordinary owning vector
(its own `str_pool`, or a retained reference to the source's), exactly as
it is when the source table was built in memory or when the `where`
keeps every row — and releasing either table in either order is safe.

## Workaround in use

Never keep a splayed-loaded table as the operand of a `select`: load into
a typed-empty in-memory table by insertion, and select over that.

```lisp
(set 'dir "/tmp/rf-splayed-str/")
(set 't (table ['id 'note] (list [1 2 3] (list "one" "two" "three"))))
(.db.splayed.set dir t)
(set 'w (table ['id 'note] (list (take [0] 0) (take [""] 0))))   ;; typed empty
(insert 'w (.db.splayed.get dir))                                 ;; copies every column
(set 'w (select {from: w where: (not (== id 1))}))               ;; fine: w owns its pool
(insert 'w (table ['id 'note] (list [9] (list "nine"))))
(print (at w 'note))                                              ;; ["two" "three" "nine"], exit 0
```

With this shape every variation in the table above is ok, including
case 1's rebind-then-insert.

ser-vasilich:
Fixes #426

## The defect

A splayed string column and its pool are written contiguously and mapped together. Every result gathered from such a column shares the pool **by reference** — filter, `take`, `drop`, all four joins and the head/tail slices reach it through `col_propagate_str_pool`, which retains it.

That retain was a promise the memory could not keep. The region belonged to the **column**: freeing the column unmapped the pool along with it, and any result still holding that pool was left reading unmapped pages. The refcount and the ownership disagreed, so taking a reference bought nothing.

**It is not a double free**, as reported. The pool is freed exactly once — by a mapping teardown that never consulted its refcount.

## Reproduces on dev, and wider than reported

Verified against `dev` (`a4776b79`), with the result bound to a name so it outlives the column:

| SIGSEGV | clean |
|---|---|
| `take`, `drop` over the column | `asc` over the column |
| `select … where` (proper subset) | `select … asc:` |
| `select … take:` | `distinct` |
| `inner-join`, `left-join`, `full-join`, `asof-join` | `select … by:` with `first` |

The split follows the code exactly: everything that shares the pool faults, everything that builds its strings afresh does not.

## The fix

The pool owns the region; the column holds an ordinary reference to it. The mapping ends when the last block living in it is freed, so `col_propagate_str_pool`'s existing retain becomes truthful and **all fifteen call sites are correct with no change to any of them**.

`mmod` grows a third value for this rather than overloading the existing "borrowed, free does nothing" — that one has other users, and conflating them would hand somebody else's mapping to this teardown. The descriptor lives on the heap because it has to outlive the region it describes.

Two ordering points, both load-bearing:

- `ray_free` reads what the teardown needs **before** releasing children, since releasing the pool can unmap the page the header itself sits on.
- `ray_release_owned_refs` now returns after releasing a string pool instead of falling through, exactly as its symbol-domain arm already did, and for the same reason. The checks after it never applied to a `STR` vec anyway.

## Tests

Nine shapes, each in its own process — the fault lands at teardown, and a process cannot observe its own. Every shape was confirmed to fault individually on the unfixed tree; the suite runner stops at the first failure in a file, so checking only the file's exit status would have proved one of them.

Two traps worth recording, both of which produced a false "clean" while writing this:

- the result must be **bound to a name**. A temporary dies before the column it came from and never reaches the dangling pool, so an inline `(print (take …))` passes even unfixed.
- `grep -x` on rayforce's vector output needs `-F`: `["one" "two"]` is a bracket expression to a regex engine.

Suite: 3713/3713 on the merged tree.

### DIFF
--- a/include/rayforce.h
+++ b/include/rayforce.h
@@ -128,6 +128,11 @@ typedef enum {
  * singleton (global intern table) or a refcounted mmapped symfile. */
 struct ray_sym_domain_s;
 
+/* One file mapping shared by more than one block (src/mem/heap.h): a mapped
+ * string column and its pool live in the same region, so neither can end it
+ * alone.  Held by the pool at aux[8..15] under mmod 3. */
+struct ray_file_map_s;
+
 typedef union ray_t {
     /* Allocated: object header */
     struct {
@@ -173,6 +178,10 @@ typedef union ray_t {
              * 8-15 hold an int64 sym ID naming the target table.
              * link_lo[8] aliases bytes 0-7.  See ops/linkop.h. */
             struct { uint8_t link_lo[8];         int64_t link_target; };
+            /* mmod 3 (string pools only): the shared mapping this block
+             * lives in.  Typed rather than reached by offset, so it moves
+             * with the union if the layout is ever reshuffled. */
+            struct { uint8_t _aux_map_lo[8];     struct ray_file_map_s* file_map; };
         };
         /* Bytes 16-31: metadata + value */
         /* Where this block's memory came from, and who ends it:
--- a/include/rayforce.h
+++ b/include/rayforce.h
@@ -175,7 +175,19 @@ typedef union ray_t {
             struct { uint8_t link_lo[8];         int64_t link_target; };
         };
         /* Bytes 16-31: metadata + value */
-        uint8_t  mmod;       /* 0=heap, 1=file-mmap */
+        /* Where this block's memory came from, and who ends it:
+         *   0  heap — the buddy allocator owns it
+         *   1  file-mmap — this block's own free munmaps the region, whose
+         *      size it derives from its type and length
+         *   2  borrowed — somebody else owns the memory and ends it on its
+         *      own schedule; free does nothing at all
+         *   3  file-mmap, shared — the region is described by a heap-side
+         *      ray_file_map_t held in aux[8..15] and ends when the last
+         *      block referencing it is freed.  A string column and its pool
+         *      live in one mapping, and a selection may keep the pool after
+         *      the column is gone, so the pool carries this and the column
+         *      merely holds a reference to it. */
+        uint8_t  mmod;
         uint8_t  order;      /* block order (block size = 2^order) */
         int8_t   type;       /* negative=atom, positive=vector, 0=LIST */
         uint8_t  attrs;      /* attribute flags */
--- a/src/mem/heap.c
+++ b/src/mem/heap.c
@@ -879,10 +879,13 @@ static void ray_release_owned_refs(ray_t* v) {
         return;
     }
 
-    /* Vector with attached index: aux[0..7] holds an owning ref to
-     * the index ray_t.  The index owns the displaced str_pool,
-     * so we must NOT also try to release those off the parent — they
-     * aren't there anymore.  Skip the STR_pool branch. */
+    /* Vector with attached index: aux[0..7] holds an owning ref to the
+     * index ray_t.  Bytes 8-15 are NOT displaced into the index for a
+     * STR or SYM parent — attach_finalize leaves the pool or domain
+     * pointer in place there (ops/idxop.c, and the layout note in
+     * rayforce.h), and ray_index_release_saved is a no-op.  So the ref
+     * taken when that pool or domain was attached is still this
+     * vector's to drop, exactly as for an unindexed one. */
     if (v->attrs & RAY_ATTR_HAS_INDEX) {
         /* A mmap-resident index (mmod==1) is a PASSENGER in this column's file
          * mapping — the column's single munmap frees it.  Releasing it here
@@ -890,6 +893,10 @@ static void ray_release_owned_refs(ray_t* v) {
          * heap-built index (mmod==0) is released by pointer. */
         if (v->index && !RAY_IS_ERR(v->index) && v->index->mmod != 1)
             ray_release(v->index);
+        if (v->type == RAY_STR && v->str_pool && !RAY_IS_ERR(v->str_pool))
+            ray_release(v->str_pool);
+        else if (v->type == RAY_SYM && v->sym_domain)
+            ray_sym_domain_release(v->sym_domain);
         return;
     }
 
@@ -1008,9 +1015,16 @@ bool ray_retain_owned_refs(ray_t* v) {
 
     if (v->attrs & RAY_ATTR_HAS_INDEX) {
         /* Mirror ray_release_owned_refs: a mmap-resident passenger index
-         * (mmod==1) is owned by the column's mapping, not refcounted here. */
+         * (mmod==1) is owned by the column's mapping, not refcounted here,
+         * and the pool or domain at bytes 8-15 is this copy's to take a
+         * ref on — skipping it is what let a copy that later sheds its
+         * index release a reference it never held. */
         if (v->index && !RAY_IS_ERR(v->index) && v->index->mmod != 1)
             ray_retain(v->index);
+        if (v->type == RAY_STR && v->str_pool && !RAY_IS_ERR(v->str_pool))
+            ray_retain(v->str_pool);
+        else if (v->type == RAY_SYM && v->sym_domain)
+            ray_sym_domain_retain(v->sym_domain);
         return true;
     }
 
--- a/src/mem/heap.c
+++ b/src/mem/heap.c
@@ -893,8 +893,14 @@ static void ray_release_owned_refs(ray_t* v) {
         return;
     }
 
-    if (v->type == RAY_STR && v->str_pool && !RAY_IS_ERR(v->str_pool))
+    if (v->type == RAY_STR && v->str_pool && !RAY_IS_ERR(v->str_pool)) {
+        /* Return rather than fall through, the way the SYM arm below does:
+         * a mapped column's pool owns the region they share, so this
+         * release can be the one that unmaps the page `v` sits on.  None
+         * of the checks after this one apply to a STR vec anyway. */
         ray_release(v->str_pool);
+        return;
+    }
 
     /* RAY_SYM vec: drop the resolution-domain ref (aux bytes 8-15).
      * No-op for the immortal runtime singleton — the common case. */
@@ -1501,6 +1507,19 @@ ray_t* ray_alloc(size_t data_size) {
  * ray_free
  * -------------------------------------------------------------------------- */
 
+/* Drop one reference to a shared file mapping; the last one unmaps it.
+ *
+ * Not atomic: every reference is taken and dropped while building or
+ * freeing a block, both of which already run under the owning heap's
+ * single-threaded discipline.  If a mapped column ever becomes shareable
+ * across heaps this needs the same treatment as ray_t's own rc. */
+void ray_file_map_release(ray_file_map_t* m) {
+    if (!m) return;
+    if (m->rc > 1) { m->rc--; return; }
+    ray_vm_unmap_file(m->base, m->len);
+    ray_sys_free(m);
+}
+
 void ray_free(ray_t* v) {
     if (!v || RAY_IS_ERR(v)) return;
     if (v->attrs & RAY_ATTR_ARENA) return;  /* arena-owned, bulk-freed */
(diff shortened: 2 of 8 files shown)


singaraiona (changes requested):
Reviewed with a verification pass — the core mechanism is sound (the pool/domain ref is now balanced across the HAS_INDEX arms, and all 15 `col_propagate_str_pool` sites become correct), but there's a **confirmed crash regression**, reproduced as an actual SIGSEGV against the PR-head build. Root cause is a single gap: the mutating column holds no reference of its own to the shared mapping, so any path that swaps or drops its `str_pool` pointer tears the region out from under live memory.

**1. Blocker — SIGSEGV in `str_pool_cow` (vec.c:1041), reproduced.** Sequence: `.db.splayed.set` then `insert` into `(get (.db.splayed.get "t/") 's)`. The temp table dies leaving the mapped STR column at rc 1; `ray_cow` returns it in place (no mmod check), `str_pool_cow`'s early-out at vec.c:1023 requires pool mmod==0 so it proceeds, deep-copies the pool, then `ray_release(vec->str_pool)` drops the mmod==3 pool 1→0 → `ray_file_map_release` 1→0 → `ray_vm_unmap_file` unmaps the entire region **including vec's own header**; the next line `vec->str_pool = new_pool` writes into the unmapped page → SIGSEGV (faulting frame confirmed at vec.c:1042 via `ray_str_vec_append`/`ray_insert`). Pre-PR the pool was mmod==2 and that release was a no-op, so this is a regression the PR introduces.

**2. Blocker — region unmapped under a live sharer, then double-munmap** (heap.c:1552/1586). The column's ownership decision keys on the *current* `str_pool` pointer, so after `str_pool_cow` swaps in a heap pool, the mmod==1 free arm reads `v->str_pool->len` from the already-freed heap pool (UAF read; overshoots `mapped_size` if the pool grew) and `ray_vm_unmap_file`s the region out from under a selection still sharing it — which later munmaps the same range a second time, potentially destroying an unrelated mapping the kernel placed there.

**3. Heap-address disclosure into a persisted file** (col.c:919). `col_save_impl` copies the live pool header and scrubs only bytes 16–31; `aux[0..15]` goes to disk verbatim, and under mmod==3 `aux[8..15]` holds the `ray_sys_alloc`'d descriptor pointer. Inert on reload (the loader overwrites aux+8) but persists across save generations via `col_copy_str_pool`. Fix: `memset(pool_header.aux, 0, 16)` after the memcpy.

The clean fix for 1, 2, and 4 at once: **give the column a real reference to the mapping** — `m->rc = 2` at map time in `col_mmap_impl` — so its free never takes the size-derived raw-unmap path and `str_pool_cow`'s release just drops rc to 1 instead of unmapping. Right now `ray_file_map_t.rc` (finding 4) is dead machinery: set to 1 once and only ever decremented, no retain helper exists, so the rc>1 branch in `ray_file_map_release` is unreachable — the exact gap findings 1–2 exploit. The comment promises "ends when the last block referencing it is freed," but only the pool ever references it.

Quality, non-blocking:
- **5.** The descriptor is stored/read via raw `memcpy` at `aux+8` (col.c:1643, heap.c:1559) with magic +8 offset and pointer size duplicated across two files, unlike the typed union members (`str_pool`, `sym_domain`, `link_target`) that share that aux slot. A forward-declared `struct ray_file_map_s* file_map;` union member makes both sites typed accesses and immune to the next aux reshuffle.
- **6.** The STR-pool/SYM-domain release+retain dispatch is copy-pasted verbatim into the HAS_INDEX arms of both `ray_release_owned_refs` (896) and `ray_retain_owned_refs` (1024), duplicating the standalone arms right below them — four hand-copied sites of exactly the drift-prone pattern whose imbalance this PR is fixing. Letting the HAS_INDEX arm fall through to the existing STR/SYM arms collapses them.

Refuted along the way (so you don't chase them): the `ray_scratch_realloc` partial-unmap variant is unreachable (mapped vecs have order 0 → effectively infinite capacity, growth branch never fires), and the non-atomic descriptor rc is moot while rc never exceeds 1 — though the `m->rc = 2` fix above changes that, so make the descriptor rc atomic if the column ref lands.

Happy to re-review once the column holds its own mapping reference — that's the one structural change the rest hangs off.

singaraiona (approved):
Re-verified on a fresh ASan/UBSan build — the structural fix is exactly right and complete. The reproduced SIGSEGV is gone: the same repro (splayed-set STR column, select subset, then insert a long string forcing `str_pool_cow`) now completes clean, exit 0, no ASan report.

Confirmed for each finding:
- **The column now holds its own reference** — `col_mmap_impl` allocates a heap-side `ray_file_map_t` with `m->rc = 2`, `pool->mmod = 3`, `pool->file_map = m`. So (1) `str_pool_cow`'s pool release drops the map rc 2→1 and returns early instead of unmapping mid-mutation, and (2) the size-derived raw-unmap path in `ray_free` (the one that UAF-read `v->str_pool->len`) is now guarded behind finding the descriptor and returning first — no UAF, no double-munmap. `ray_file_map_t.rc` is live machinery now, and the `rc>1` early-return is genuinely reached. The rc decrements go through `ray_atomic_dec` (RELEASE), increments are single-threaded at map time — sound.
- **The disclosure (finding 3) is fixed too** — `col_save_impl` now `memset(pool_header.aux, 0, 16)` before writing the header, so the `file_map` pointer no longer reaches disk.
- **Region unmapped exactly once** — traced all drop orderings (column-first, selection-first, post-COW) under ASan/LSan: no leak, no double-free, munmap happens once when the last of {pool, column, sharers} goes.

Suite: 3716/3716 under ASan/UBSan, and all three added tests pass (`splay_str_column_roundtrip` asserting `mmod == 3`, `indexed_col_aux_refs`, and the `splayed_str_pool_lifetime` growth test). The one UBSan notice in the log (`test_aof.c:85`) is pre-existing and unrelated — this PR doesn't touch that file.

One non-blocking nit: the block comment above `ray_file_map_release` still says "Not atomic" while the struct comment says "Atomic" and the code uses the atomic dec — worth tidying the stale comment.

Nice fix — giving the column its own hold was the clean way through. Approving and merging.

Issue #414, opened 2026-08-17
select: virtual row index i raises a bare nyi when where: and by: are combined, and in any where: clause


vbanas:
# select: virtual row index `i` raises a bare `nyi` when `where:` and `by:` are combined, and in any `where:` clause

**Engine:** `v2.5.13-88-g212fff3a` (dev, commit `212fff3a`), Linux x86-64.

The virtual row-index column `i` is accepted in some `select` shapes and
raises in others — and the failing shapes include the most idiomatic
row-count query there is, `(count i)` per group over a filtered table.

## Repro

With nothing loaded but a four-row table:

```clj
(set t (table ['g 'd 'v] (list (list 'a 'a 'b 'b)
                               (list 2026.01.01 2026.01.02 2026.01.01 2026.01.02)
                               (list 10 20 30 40))))

(select {n: (count i) from: t by: {k: d}})                     ; ok
(select {n: (count i) from: t where: (== g 'a)})               ; ok
(select {n: (count i) from: t where: (== g 'a) by: {k: d}})    ; error: nyi
```

Either clause alone works; the combination raises. The `by:` form does
not matter (brace `{k: d}`, bare `d`, list `[d]` — all raise), and it
raises whether `(count i)` is the only aggregate or one of several.
`(max i)` and `(first i)` in the same combined shape both work, so the
gap is specific to `count` and to bare projection — not to `i` under
grouping in general. Separately, `i` inside the `where:` clause raises
even without `by:`, so `i` cannot be used to filter by row position at
all:

```clj
(select {n: i from: t where: (== g 'a) by: {k: d}})            ; error: nyi
(select {from: t where: (< i 2)})                              ; error: nyi
```

| query shape | result |
|---|---|
| `(count i)` + `by:` only | ok |
| `(count i)` + `where:` only | ok |
| `(count i)` + `where:` + `by:` | **error: nyi** |
| `(count <real col>)` + `where:` + `by:` | ok |
| `(max i)` / `(first i)` + `where:` + `by:` | ok |
| bare `i` output + `where:` + `by:` | **error: nyi** |
| `i` inside `where:`, with or without `by:` | **error: nyi** |

## Probe script and its output

```clj
(set t (table ['g 'd 'v] (list (list 'a 'a 'b 'b)
                               (list 2026.01.01 2026.01.02 2026.01.01 2026.01.02)
                               (list 10 20 30 40))))
(set probe
  (fn [label form]
[...]

PR #437, opened 2026-08-26
fix(query): close the two ways an unresolved name still passed a select

### LINKED ISSUE #428: Engine bug — an unknown column name inside `select` resolves to nothing instead of raising, and the surviving projections shift under the wrong labels
Found 2026-08-24 at engine `v2.5.15` (`605723d6`),
the pinned release; not re-tested on `dev`.

---

**Title:** fix(query): a bare column name that the operand table does not have
is silently dropped inside `select` — projections shift left under the
original labels, and aggregates answer a neutral value

## Summary

Inside a `select`, a bare symbol that names no column of the operand table
does not raise. It resolves to nothing, and what happens next depends on
where it sat:

1. **As a projection value** the `name: value` pair is dropped from the
   VALUE list while the NAME list keeps its length, so the remaining values
   shift left and are emitted under the **wrong labels**. The result is a
   narrower table whose columns look right and hold another column's data.
2. **Under an aggregate** the aggregate is applied to nothing and answers its
   neutral element — `sum` → `0`, `count` → the operand's row count, `max` →
   `INT64_MIN`, `first` → `0`.

Neither case prints or returns any diagnostic. The same name evaluated
outside a `select` raises `name`, and the same name wrapped in any
expression inside a `select` (`(+ nosuch 1)`) also raises — so the silence
is specific to a *bare* column reference, which is exactly the form every
ordinary projection uses.

`where:` and `by:` clauses do raise on an unknown name, and so does a bare
unknown projection in a select that carries `by:`. Only the ungrouped
projection list and the aggregate arguments are affected.

The consequence is that a typo, a renamed column, or a schema drift turns a
query that should fail loudly into one that returns a plausible table of
wrong numbers.

## Environment

- rayforce `v2.5.15` (`605723d6b44d95d41244c9abd27fa99bae0d1ec6`), the
  `rayforce` binary built from the release tag.
- Linux x86_64 (Ubuntu, kernel 7.0.0-28-generic), glibc 2.39.

## Minimal reproducer

`repro.rfl`:

```lisp
(set t (table ['a 'b 'c] (list [1 2 3] [10 20 30] [100 200 300])))

;; 1. `nosuch` is not a column of t. Expected: an error naming it.
(show (select {x: a y: nosuch z: c from: t}))

;; 2. an aggregate over the same name
(show (select {s: (sum nosuch) from: t}))

;; 3. the same name outside a select — this DOES raise
(println nosuch)
```

```text
$ rayforce repro.rfl
┌─────┬─────┐
│  x  │  y  │
│ I64 │ I64 │
├─────┼─────┤
│ 1   │ 100 │
│ 2   │ 200 │
│ 3   │ 300 │
└─────┴─────┘
┌─────┐
│  s  │
│ I64 │
├─────┤
│ 0   │
└─────┘
error: name: 'nosuch' undefined
```

The first table is the bug in one picture: two columns instead of three,
and the column **labelled `y` holds `c`**. A caller who asked for
`{x: a  y: nosuch  z: c}` and reads `y` gets `c`'s values with no
indication that anything went wrong.

## What was varied, and what happened

Operand `t` has columns `a`, `b`, `c`; `g` is `(table ['k 'a] …)`.
`nosuch` / `alsonot` are bound nowhere.

| variation | result |
|---|---|
| `{x: a  y: b  z: c}` (all known) | `[x y z]`, correct — the control |
| `{x: a  y: b  z: nosuch}` (unknown LAST) | `[x y]`; `y` holds `b` — right data, one column silently missing |
| `{x: a  y: nosuch  z: c}` (unknown MIDDLE) | `[x y]`; `y` holds **`c`** — mislabelled |
| `{x: nosuch  y: b  z: c}` (unknown FIRST) | `[x y]`; `x` holds **`b`**, `y` holds **`c`** — both mislabelled |
| `{x: nosuch  y: alsonot}` (all unknown) | `[]` — a table with no columns |
| `{s: (sum nosuch)}` | `0` |
| `{s: (count nosuch)}` | `3` — the operand's row count |
| `{s: (max nosuch)}` | `-9223372036854775808` (`INT64_MIN`) |
| `{s: (first nosuch)}` | `0` |
| `{n: (count i)  s: (sum nosuch)  from: g  by: k}` | `s` is `0` in every group |
| `{n: (count i)  z: nosuch  from: g  by: k}` | **raises** |
| `{from: t  where: (> nosuch 1)}` | **raises** |
| `{n: (count i)  from: t  by: nosuch}` | **raises** |
| `{x: (+ nosuch 1)  from: t}` | **raises** |
| `nosuch` evaluated outside any `select` | **raises** `name` |

So the engine already has the right behaviour in five of these positions.
The gap is the bare projection value and the aggregate argument.

## Expected

A bare name inside `select` that is neither a column of the operand nor a
binding in scope raises the same `name` error it raises everywhere else,
naming the symbol. Failing that, at minimum the projection's name list and
value list stay aligned, so a surviving column can never carry another
column's data under its own label.

## Why the shift matters more than the omission

Dropping a column a caller asked for is recoverable — the caller sees a
narrower table. Emitting `c`'s values under the label `y` is not: every
downstream consumer, including a human reading a rendered table, has no
signal at all. In the "unknown FIRST" row above, *every* column of the
result is mislabelled while the table still looks entirely ordinary.

## Workaround in use

None that the language enforces. Callers that build projection lists
dynamically check the name against `(cols t)` before putting it in a
`select`; hand-written queries have nothing but review. This is the reason
the bug is worth fixing rather than documenting: there is no shape a caller
can adopt that makes the silent case loud.

ser-vasilich:
Fixes #428

## Already fixed, except for two ways around it

The report was filed against `v2.5.15` and describes the same defect as #414, which has since been fixed: an unknown column reference now gets an invalid-scan marker that `ray_execute` rejects. Every shape in the report raises `schema: column 'nosuch' not found` on current `dev`.

Two ways around that marker survived, and both reproduce the report in full — a projection emitted under the wrong label, and an aggregate answering its neutral element. Neither should close as a duplicate.

## Hole A — the `_dist` exemption was unconditional

`_dist` is exempt from the marker because a rerank produces that column *while the query runs*, so it is genuinely absent when the query compiles. But the exemption applied just as well to a query with no `nearest:` clause, where nothing will ever produce it. Verified on `dev` before the change:

```
(select {x: a d: _dist z: b from: T})  ->  a two-column table {x, d}
                                           where d holds b's data, z dropped
(select {s: (sum   _dist) from: T})    ->  0
(select {c: (count _dist) from: T})    ->  the operand's row count
(select {m: (max   _dist) from: T})    ->  INT64_MIN
```

The exemption is now conditional on the graph actually holding an `OP_ANN_RERANK` / `OP_KNN_RERANK` node. The rerank is built before projections compile, so the check is a walk over nodes already present, on the cold unknown-name path only.

## Hole B — `nearest:` missing from the rename loop's skip list

The loop that names output columns is positional, so a clause keyword it fails to skip consumes an output slot and every name after it lands one column early. `nearest:` is counted out where the outputs are counted but was not skipped here, so a query naming it before its outputs labelled them one position off:

```
(select {from: TN nearest: (knn Vec [1.0 0.0]) i: id d: _dist take: 2})
  ->  columns labelled  nearest | i        instead of  i | d
```

## Not changed

The marker-plus-`validate_scan_columns` split stays as it is. Raising at resolution time was considered and rejected for the reason the existing comment records: optimizer rewrites can legitimately make a scan dead, and an eager error would have to be threaded through every `compile_expr_dag` caller's NULL contract.

## Tests

`test/rfl/regress/issue_428.rfl` covers both holes and the legitimate `_dist` case that must keep working. Confirmed red before the change: the projection shape returned a two-column table whose second column carried the third column's data under the second column's name.

No existing test depended on the silent behaviour — the assertions in this area all *demand* the error.

Suite: 3713/3713.

### DIFF
--- a/src/ops/query.c
+++ b/src/ops/query.c
@@ -1128,8 +1128,22 @@ ray_op_t* compile_expr_dag(ray_graph_t* g, ray_t* expr) {
          * shortcuts may make the scan dead (count/max/projection paths), but
          * execution must still reject the original column reference. */
         ray_op_t* missing = ray_scan(g, ray_str_ptr(s));
+        /* `_dist` is exempt, but only where it can legitimately appear: a
+         * rerank produces it during execution, so it is absent from the
+         * operand table at compile time.  Exempting it unconditionally let
+         * every symptom of an unknown column back in for that one name —
+         * a projection under `_dist` shifted the labels left, and an
+         * aggregate over it answered its neutral element — in queries with
+         * no nearest: clause at all. */
         int64_t dist_sym = ray_sym_find("_dist", 5);
-        if (missing && expr->i64 != dist_sym) {
+        bool exempt = false;
+        if (missing && expr->i64 == dist_sym) {
+            for (uint32_t i = 0; i < g->node_count && !exempt; i++)
+                if (g->nodes[i].opcode == OP_ANN_RERANK ||
+                    g->nodes[i].opcode == OP_KNN_RERANK)
+                    exempt = true;
+        }
+        if (missing && !exempt) {
             missing->flags |= OP_FLAG_INVALID_SCAN;
             ray_op_ext_t* ext = find_ext(g, missing->id);
             if (ext) ext->base.flags |= OP_FLAG_INVALID_SCAN;
@@ -10154,8 +10168,16 @@ by_dict_done:
                 int j = 0;
                 for (int64_t i = 0; i + 1 < dict_n; i += 2) {
                     int64_t kid = dict_elems[i]->i64;
+                    /* Every clause keyword has to be skipped here, not just
+                     * most of them: this walk is positional, so a keyword it
+                     * fails to skip consumes an output slot and every name
+                     * after it lands on the wrong column.  nearest: was
+                     * missing, and it is counted out where the outputs are
+                     * counted, so a query naming it before its outputs
+                     * labelled them one position early. */
                     if (kid == from_id || kid == where_id || kid == by_id ||
-                        kid == take_id || kid == asc_id || kid == desc_id) continue;
+                        kid == take_id || kid == asc_id || kid == desc_id ||
+                        kid == nearest_id) continue;
                     if (n_key_cols + j < ncols)
                         ray_table_set_col_name(result, n_key_cols + j, kid);
                     j++;
--- a/test/rfl/regress/issue_428.rfl
+++ b/test/rfl/regress/issue_428.rfl
@@ -1,0 +1,43 @@
+;; A name a select cannot resolve must raise, and every clause keyword must
+;; be skipped when the outputs are named.
+;;
+;; The general case was fixed with #414: an unknown column reference gets an
+;; invalid-scan marker that ray_execute rejects.  Two ways around it
+;; survived, and both produced the #428 report — a projection under the
+;; wrong label, and an aggregate answering its neutral element.
+;;
+;; 1. `_dist` was exempt from the marker unconditionally.  The exemption is
+;;    needed, because a rerank produces that column during execution and it
+;;    is genuinely absent when the query compiles — but it applied just as
+;;    well to a query with no nearest: clause, where nothing will ever
+;;    produce it.
+;;
+;; 2. `nearest:` was missing from the skip list in the loop that names the
+;;    output columns.  That walk is positional, so a keyword it fails to
+;;    skip eats an output slot and every name after it lands one column
+;;    early.  It is counted out where the outputs are counted, which is why
+;;    only the naming disagreed.
+
+(set _T (table [a b] (list [1 2 3] [4 5 6])))
+
+;; ── 1. `_dist` with no rerank to produce it: an ordinary unknown name ────
+(select {x: a d: _dist z: b from: _T})        !- schema
+(select {d: _dist from: _T})                  !- schema
+(select {s: (sum _dist) from: _T})            !- schema
+(select {c: (count _dist) from: _T})          !- schema
+(select {m: (max _dist) from: _T})            !- schema
+(select {f: (first _dist) from: _T})          !- schema
+(select {s: (sum _dist) by: a from: _T})      !- schema
+;; and the plain unknown name it used to be exempt from behaving like
+(select {x: a y: nosuch z: b from: _T})       !- schema
+
+;; ── 2. `_dist` where a rerank does produce it — still resolves, still
+;;    correct.  The nearest row is id 1 at distance 0. ───────────────────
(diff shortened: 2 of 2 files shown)


singaraiona (approved):
Reviewed with an empirical verification pass (head + base built ASan/UBSan, repros run on both). The fix is correct and does exactly what it says — both escape routes for an unresolved name are closed: the `_dist` exemption is now gated on a rerank node actually being present, and `nearest_id` is added to the output-naming skip list so it matches the output-count loop and `select_output_count` exactly. Suite 3713/3713 clean. Approving.

Verified, so reviewers don't re-chase them:
- **`where: _dist` is not a regression.** `(select {i: id from: TN where: (< _dist 0.5) nearest: (knn Vec [1.0 0.0]) take: 2})` fails with a schema error on *both* head and base — the WHERE reference compiles before the rerank node exists, so the exemption legitimately can't fire there. Head just gives a more specific message. Not newly broken.
- The keyword sets now agree exactly, the keyword universe is closed at seven, and every other positional dict-walk that still omits `nearest_id` (7056/7341/7614/8147/8575/8818/2480/2561) is inside a `by:`-only path, unreachable today because `nearest:` + `by:` is rejected up front at query.c:5727.

One optional hardening (non-blocking): the regression test's `[1 3]` assertion rests on an exact floating-point tie — id 3 `[1.0 1.0]` and id 4 `[0.5 0.5]` have identical cosine distance to q=[1,0], and the result is decided only by the heap keeping the earlier-inserted row. On this platform the two distances are bit-identical at -O0 and -O2 (both reduce to correctly-rounded `sqrt(0.5)`, so it's actually not FMA/libm-dependent for these inputs), so it passes robustly — but the test asserts a golden that rests on a tie rather than a real ordering, which is needlessly fragile for the next contributor. Changing id 4's vector to make the distances distinct (e.g. `[0.5 0.5]` → `[0.9 0.1]`) exercises the identical path and passes with zero golden edits — verified.

Two pre-existing follow-ups surfaced while tracing (neither introduced by this PR, both behave identically on base — worth a separate issue, not a blocker here):
- The projection eval-fallback silently drops `nearest:` when there's no `where:`: `(select {u: (distinct id) from: TN nearest: (knn Vec [1.0 0.0]) take: 2})` returns table-order rows, not nearest, with no error (confirmed same on head and base). `distinct` doesn't DAG-lower, so the fallback frees the graph including the rerank.
- The exemption keys on graph *state* rather than the query — it works because outputs compile after the rerank node, but that coupling to compile order isn't enforced by anything. `nearest_expr != NULL` (already computed at query.c:5690) would be order-independent; worth considering if this area is touched again.

Approving and merging — I'll fold the test-tie tweak suggestion in as a comment rather than blocking on it.

PR #443, opened 2026-08-30
fix(syscmd): flush REPL help output

vbmithr:
## Summary

- flush stdout after rendering the REPL command list
- make `:help` immediately visible to pipe-backed clients such as Emacs Comint

## Validation

- `make debug`
- started Rayforce with a pipe-backed Emacs process, sent `:help`, and verified that `Commands list` arrived before EOF

### DIFF
--- a/src/app/repl.c
+++ b/src/app/repl.c
@@ -920,6 +920,11 @@ static bool handle_command(ray_repl_t* repl, const char* str, size_t len) {
         ray_release(result);
     }
 
+    /* A piped REPL uses block-buffered stdout.  Every command response is
+     * complete at this boundary, including output printed by handlers and
+     * errors printed above, so make it visible before waiting for more input. */
+    fflush(stdout);
+
     /* Track timeit echoing so the prompt reflects current state. */
     repl->timeit = g_ray_profile.active;
     return true;


vbmithr:
`h_env` probably needs it too?

singaraiona (commented):
The flush itself is correct and harmless in interactive mode, and the premise is real: in pipe mode `handle_command` is invoked with no flush afterward, so `:h` output sits in the block buffer until a later evaluation flushes.

But the same bug remains in every sibling path: `h_timeit` (`syscmd.c:110-115`) and `h_env` (`syscmd.c:174-181`) print to stdout in REPL mode and return without flushing, and `handle_command` itself prints the unknown-command hint and error messages (`repl.c:850-863`) unflushed. Drive the REPL over a pipe and type `:t`, `:env`, or a typo like `:foo` — same invisible-output symptom this PR fixes, for three of the four affected commands.

Suggest moving the fix one level up: a single `fflush(stdout)` at the end of `handle_command` in `repl.c` (the shared syscmd boundary) covers all of them and makes the `h_help`-local flush unnecessary.

singaraiona:
No extra flush needed — this revision already covers `h_env`. The handlers only print to stdout when `ctx->repl` is set (`h_env` under `.sys.cmd` gets `ctx=NULL` and returns the count instead of printing), and the only dispatch site that passes a REPL context is `handle_command`, so every REPL-mode print — `:h`, `:t`, `:env`, and the error/unknown-command messages — now drains through the single `fflush(stdout)` at the end of `handle_command`. This revision looks right to me.

PR #446, opened 2026-08-31
fix(mem): hand a cross-thread free back to its owner, and return idle pages to the OS

### LINKED ISSUE #439: Composite `by:` group-by leaks the hash table on every execution (v2.5.14)
## Summary

A `select` with a **composite** `by:` key leaks memory proportional to the number
of groups, on every execution. The memory is never returned, so running the same
query repeatedly in one process grows RSS linearly until the process OOMs.

A single-column `by:` over the *same* number of groups does not leak, so this is
specific to the multi-key group path rather than to grouping in general.

Found while benchmarking Rayforce on ClickBench
([ClickHouse/ClickBench#1375](https://github.com/ClickHouse/ClickBench/pull/1375)),
where it is the reason several queries degrade or kill the server.

## Repro

`leak.rfl` — 20M rows, 20M distinct `[a b]` pairs, the same query six times:

```lisp
(set n 20000000)
(set idx (til n))
(.db.splayed.set "t" (table [a b] (list idx (% idx 1000))))
(set t (.db.splayed.get "t"))

(set r (select {from: t by: [a b] c: (count a) desc: c take: 10}))
(set r (select {from: t by: [a b] c: (count a) desc: c take: 10}))
(set r (select {from: t by: [a b] c: (count a) desc: c take: 10}))
(set r (select {from: t by: [a b] c: (count a) desc: c take: 10}))
(set r (select {from: t by: [a b] c: (count a) desc: c take: 10}))
(set r (select {from: t by: [a b] c: (count a) desc: c take: 10}))
(exit 0)
```

```
$ /usr/bin/time -v rayforce leak.rfl
```

Only one result is live at a time (each `set r` drops the previous one), and the
result is 10 rows.

## Numbers

Peak RSS as a function of how many times the identical query runs in one process:

| repeats | `by: [a b]` (20M groups) | `by: a` (20M groups) | `by: b` (1000 groups) | no grouping |
|--:|--:|--:|--:|--:|
| 1 | 6.6 GB | 0.80 GB | 0.17 GB | 0.17 GB |
| 2 | 9.3 GB | 0.80 GB | 0.17 GB | — |
| 3 | 12.5 GB | 0.79 GB | 0.17 GB | — |
| 6 | 22.7 GB | 0.80 GB | 0.17 GB | 0.17 GB |

`by: [a b]` grows by ~3.2 GB per execution and never comes back down. The other
three are flat. Note the second column: same 20M groups, same aggregate, single
key — no leak. Nothing here is retained by the script.

It scales with the group count, not the row count:

| rows / groups | peak after 1 | peak after 4 | leaked per execution |
|--:|--:|--:|--:|
| 5M | 2.2 GB | 5.7 GB | 1.16 GB (~233 B/group) |
| 10M | 3.7 GB | 9.5 GB | 1.90 GB (~195 B/group) |
| 20M | 6.6 GB | 16.0 GB | 3.09 GB (~158 B/group) |

Against a server (`rayforce -p 5000`) the same thing happens across IPC requests,
so a long-lived server accumulates it until queries start failing:

```
try 1: 2.83 s  RSS 4.5 GB
try 2: 1.86 s  RSS 7.9 GB
try 3: 2.60 s  RSS 11.2 GB
try 4: 3.49 s  RSS 14.4 GB
try 5: 1.82 s  RSS 17.6 GB
try 6: 1.79 s  RSS 20.8 GB
```

A cheap query issued afterwards runs fine and frees nothing (RSS stays at
20.8 GB), so it is not the last result being held.

`-m` does not contain it. With `-m 200M` RSS still climbed to 19 GB over five
executions before `error: oom`, and the spill path made one execution take
101 s instead of ~1 s.

## Why it matters

On ClickBench (100M rows, 43 queries, 3 runs each) the queries with composite
keys and ~100M groups are exactly the ones that misbehave, and the extrapolated
leak (~16 GB per execution at 100M groups) matches what happens on a 32 GB
machine:

* `by: [MobilePhone MobilePhoneModel]` with `(count (distinct UserID))` — server
  dies during the first run (`error: io: connection closed`)
* `by: [UserID SearchPhrase]` — run 1 is 4.30 s, runs 2 and 3 are 2.42 s and
  52.07 s; a second full benchmark on an identical machine got 11.39 s and
  92.71 s for the same two runs
* `by: {UserID (minute EventTime) SearchPhrase}` — 8.65 s, then 54.51 s and
  62.03 s
* `by: [WatchID ClientIP]` — 88 s, then `error: oom` twice

Run 1 of each query is reproducible to three significant digits across
independent benchmark runs, because the benchmark restarts the server before
every query. Runs 2 and 3 are not, because they execute in a process the first
run has already poisoned.

## Environment

* Rayforce v2.5.14 (`6067f3a`), built from source with `make release`
* Linux aarch64 (also observed on x86-64 — the ClickBench runs above are
  `c6a.4xlarge` and `c8g.4xlarge`)

ser-vasilich:
Closes #439.

## The bug

A repeated `select` with a composite `by:` key grew a served process on every execution and never gave any of it back, until `RAY_MAX_POOLS` turned it into an `oom`.

Nothing leaked. Workers allocate their per-group temporaries from their own heaps, and the thread that consumes the result frees them. A block that is not in any of the freeing heap's pools took the "foreign" path: `heap_find_pool` misses, absorbing it locally would divorce it from the pool that owns it, so it was queued on **the freeing thread's own** list to wait for a collector. `repl.c` runs one after every statement; `ipc.c` ran none at all. So a served query parked its whole allocation volume on the connection thread, the owning worker began the next execution short by exactly that much, and mapped a fresh pool to cover it.

## Hand the block to its owner instead

`ray_free` now resolves the owner from the pool header and CAS-pushes the block onto **that heap's** list; the owner drains it with one atomic exchange in `ray_alloc`'s slow path, before any pool can be added. Nothing accumulates anywhere, so the growth is impossible by construction and no collector call is involved.

That needs a lifetime guarantee the allocator did not have: a heap must outlive every block it owns, or a freeing thread can CAS into a mapping the owner has already unmapped. So a worker now **abandons** its heap — it stays registered with its pools intact, and the next worker thread adopts it — instead of destroying it. Real teardown is kept for process shutdown.

Four mechanisms existed only to work around the old accumulate-locally design and are gone: returning foreign blocks during collection, absorbing them onto a local freelist, the registry-wide `has_foreign` scan in the pool-reclaim pass together with the destructive purge that had to follow it, and the cross-heap purge in `ray_heap_destroy` — and with it the freelist-tearing hazard its own comments documented. The dead `heap_flush_foreign(h, false)` call in `ray_alloc` is removed.

A swap-backed pool's file is now unlinked as soon as it is mapped rather than at teardown, which an abandoned heap never reaches. The mapping holds the inode, so the blocks stay reserved while the pool is in use and the kernel reclaims them on unmap.

## Return the pages once the process goes quiet

The heaps stop growing, but a mapped pool stays mapped, so a process that ran one heavy query keeps its peak resident for good. There is no thread and no queue: a unit of work stamps a coarse timestamp, and the points that already exist — the statement boundary, and the poll loop, which bounds its wait by the deadline exactly as it already bounds it by the next timer — check whether enough time has passed and, if so, walk every heap's free blocks and hand their pages back.

The timestamp is the whole design. Releasing pages as soon as a lot becomes free is wrong: for a repeated query those are precisely the pages the next iteration faults straight back in. Only elapsed idleness distinguishes "the workload has stopped" from "we are between two iterations of a hot loop". The threshold defaults to 10s and `RAY_HEAP_DECAY_MS` overrides it; negative disables it.

Crossing heaps is sound here for the same reason it is in the existing passes: `ray_parallel_end` drains `pending` with acquire before clearing `ray_parallel_flag`, and a worker allocates nothing between its last `pending--` and `sem_wait`, so a zero flag means every worker heap is quiescent. A concurrent cross-thread free can only push, which the atomic-exchange drain handles.

`ray_heap_release_pages` keeps its synchronous contract for the fallback in `join.c`, which fires only after `scratch_alloc` has already failed — under proven pressure, refaulting beats an `oom`.

## Tests

Regression tests for the owner-side free and for the decay, each failing on its parent commit. The decay tests assert both halves — that a collection does **not** release an aged block and that the decay immediately after does, and that nothing is swept before the threshold — and each asserts that it covered something, so neither can pass vacuously.

### DIFF
--- a/src/app/repl.c
+++ b/src/app/repl.c
@@ -813,11 +813,6 @@ static void eval_and_print(ray_term_t* term, const char* input,
     ray_qlog_ctx_t qc;
     ray_qlog_begin(&qc);
 
-    /* Stamp the allocator's idle clock at the START of the statement, so
-     * the decay check after it measures how long we were quiet BEFORE this
-     * line rather than how long the line took. */
-    ray_heap_note_activity();
-
     ray_term_clear_interrupt();
     ray_eval_clear_interrupt();
     if (term) ray_term_eval_begin(term);
@@ -879,11 +874,15 @@ static void eval_and_print(ray_term_t* term, const char* input,
 
     if (profiling) profile_print(use_color);
     ray_heap_gc();
-    /* Statement boundary: if the gap before this line already exceeded the
-     * decay threshold, hand the pages back now.  In a loop of short
-     * statements the gap is the statement's own duration and this is a
-     * timestamp comparison that does nothing. */
+    /* Statement boundary.  Check BEFORE stamping: the clock still holds the
+     * end of the previous statement, so what we measure is the gap between
+     * statements — the time the process actually sat idle.  Stamping first
+     * would measure this statement's own duration instead, and a loop of
+     * statements longer than the threshold would then sweep after every one
+     * of them, discarding the working set the next is about to fault back
+     * in. */
     ray_heap_decay();
+    ray_heap_note_activity();
 }
 
 /* `type_label` and `cmd_match` were inlined into the previous bespoke
--- a/src/app/repl.c
+++ b/src/app/repl.c
@@ -1394,6 +1394,15 @@ int ray_repl_run_file(const char* path) {
 
     /* profile tree goes to stdout via profile_print; honour stdout's tty. */
     if (profiling) profile_print(color_out);
+    /* A whole file is one unit of work, and unlike the REPL there is no
+     * decay check after it — so stamp at the END, not the start.  A script
+     * that exits has nothing to give back; a script that starts a server
+     * and hands over to the poll loop leaves its load-time temporaries
+     * resident, and this is what lets the loop notice them.  Stamping at
+     * the start would instead arm a check that the script's own (.sys.gc)
+     * calls would service mid-run, which is the one thing the threshold
+     * exists to avoid. */
+    ray_heap_note_activity();
     ray_heap_gc();
     return rc;
 }
--- a/src/app/repl.c
+++ b/src/app/repl.c
@@ -813,6 +813,11 @@ static void eval_and_print(ray_term_t* term, const char* input,
     ray_qlog_ctx_t qc;
     ray_qlog_begin(&qc);
 
+    /* Stamp the allocator's idle clock at the START of the statement, so
+     * the decay check after it measures how long we were quiet BEFORE this
+     * line rather than how long the line took. */
+    ray_heap_note_activity();
+
     ray_term_clear_interrupt();
     ray_eval_clear_interrupt();
     if (term) ray_term_eval_begin(term);
@@ -874,6 +879,11 @@ static void eval_and_print(ray_term_t* term, const char* input,
 
     if (profiling) profile_print(use_color);
     ray_heap_gc();
+    /* Statement boundary: if the gap before this line already exceeded the
+     * decay threshold, hand the pages back now.  In a loop of short
+     * statements the gap is the statement's own duration and this is a
+     * timestamp comparison that does nothing. */
+    ray_heap_decay();
 }
 
 /* `type_label` and `cmd_match` were inlined into the previous bespoke
--- a/src/core/epoll.c
+++ b/src/core/epoll.c
@@ -26,6 +26,7 @@
 #include "core/poll.h"
 #include "core/timer.h"
 #include "mem/sys.h"
+#include "mem/heap.h"   /* idle decay: bound the wait, sweep after wakeup */
 #include <sys/epoll.h>
 #include <unistd.h>
 #include <errno.h>
@@ -186,6 +187,18 @@ int64_t ray_poll_run_for(ray_poll_t* poll, int timeout_ms)
             }
         }
 
+        /* A process waiting here is a process doing nothing, and this is the
+         * only moment it gets to notice.  Bound the wait by the allocator's
+         * pending decay exactly as a timer would; once the decay has run
+         * there is nothing pending and the wait goes back to indefinite. */
+        {
+            int64_t decay = ray_heap_decay_due_ms();
+            if (decay >= 0) {
+                if (decay > INT_MAX) decay = INT_MAX;
+                if (wait_ms < 0 || decay < wait_ms) wait_ms = (int)decay;
+            }
+        }
+
         int n = epoll_wait((int)poll->fd, events, RAY_POLL_MAX_EVENTS, wait_ms);
         if (n < 0) {
             if (errno == EINTR) continue;
@@ -274,6 +287,10 @@ int64_t ray_poll_run_for(ray_poll_t* poll, int timeout_ms)
 
         if (poll->timers)
             ray_timers_fire_expired((ray_timers_t*)poll->timers);
+        /* After the events of this wakeup, not before: a request handled
+         * above has just re-stamped the activity clock, so a busy loop
+         * finds nothing due and pays one timestamp comparison. */
+        ray_heap_decay();
         if (bounded) break;
     }
 
--- a/src/core/epoll.c
+++ b/src/core/epoll.c
@@ -290,8 +290,14 @@ int64_t ray_poll_run_for(ray_poll_t* poll, int timeout_ms)
         next_event:;
         }
 
-        if (poll->timers)
-            ray_timers_fire_expired((ray_timers_t*)poll->timers);
+        if (poll->timers) {
+            /* A fired timer is work, and on a timer-driven server it is the
+             * only work there is: without this stamp the decay sees an idle
+             * process mid-workload, and — having disarmed itself — never
+             * looks again. */
+            if (ray_timers_fire_expired((ray_timers_t*)poll->timers) > 0)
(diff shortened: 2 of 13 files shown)


singaraiona (changes requested):
CI is green (incl. TSAN) and the core owner-routed free design looks sound — the deep review refuted the scary candidates (foreign-LIFO coalesce corruption, epoll busy-spin, registry data race, pass-4 lock removal). But two decay-policy bugs are real and should be fixed before merge, plus one project-rule issue:

**1. Activity is stamped at the START of a statement but checked at the END** (`repl.c:819`, and `ipc.c:904` at frame arrival). The measured "quiet gap" is the statement's own runtime, so a loop of 15s queries (threshold 10s) fires a full un-aged, un-budgeted sweep at every statement boundary — releasing exactly the warm working set the next iteration refaults, in perpetuity. The comment at `repl.c:816-818` ("measures how long we were quiet BEFORE this line") is wrong; the script path at `repl.c:1405` already stamps at the END with a comment explaining why. Stamp at the end of the unit of work (or check due-ness before the start-stamp).

**2. Timer callbacks never stamp activity** (`ray_timers_fire_expired` → `call_fn1`, no `ray_heap_note_activity`). A timer-driven server (a) gets the sweep fired between two timer fires mid-workload, and (b) after that one sweep `g_heap_decay_armed` stays false forever, so later-accumulated footprint never decays even across genuine idle. The comment at `epoll.c:295-297` / `kqueue.c:297-299` ("a request handled above has just re-stamped the activity clock") is false for timer-only wakeups. One stamp in the poll loop when timers fired fixes both.

**3. `RAY_HEAP_DECAY_MS` is a new ad-hoc env knob** — against the project rule (tunables derive from existing budgets or go through existing surfaces; cf. `RAY_DIRECT_CACHE_MB`, removed in 1ae57b98). It also re-runs `getenv`+`strtoll` on every check, twice per poll wakeup, for a value that can't change after exec. Parse once into a static, give the tests a public setter (the `ray_heap_direct_cache_drain` precedent), and prefer deriving the default/surface from existing budget machinery.

Confirmed-but-latent (public API reachable only from tests today — fix or fence):
- `ray_heap_merge` drains `src->foreign` without decrementing `bytes_allocated` per block — under the new owner-routed model those are src-owned blocks still in src's stats, so `.sys.mem` drifts up permanently on every merged pending block. One-line fix in the drain loop.
- `ray_free`'s owner-gone early return silently drops the block, but a heap unregistered via `ray_heap_push_pending` still has its pools mapped and later merged — a cross-thread free in that window leaks the block and pins its pool against pass-4 reclamation forever. The comment at the return ("nothing to leak") is false on this path.

Cleanups worth doing while in here:
- `_Atomic foreign` sits at offset 16 of `ray_heap_t`, sharing the owner's hottest cache line (`avail`, `slabs[0]`); the new model makes it a remote CAS target, so consumer threads bounce the owner's alloc/free fast-path line on every push. `_Alignas(64)` it onto its own line.
- The decay walk is the third hand-synced copy of the release+mark loop (gc pass 5, `ray_heap_release_pages`, decay), with the order-13 floor as a bare literal in four places — extract `heap_release_free_pages()` and name the floor next to `RAY_FREE_AGE_RELEASE`.
- The rewritten gc pass-3 comment re-asserts the stale "workers may still be calling ray_free between pending-- and sem_wait" claim that the decay sweep's own justification (`heap.c:2482-2489`) refutes; state the real (policy) reason instead.

singaraiona (commented):
Re-reviewed 9778f14b, 5a7e6bd9, 4190e012 — the substantive findings are all properly addressed: end-of-work stamping with check-before-stamp, the timer stamp in both poll loops, merge accounting, the owner-gone window, `_Alignas(64)` on `foreign`, the shared release walk, and the pass-3 comment. Nice turnaround.

Two leftovers before this merges:

1. **The `RAY_HEAP_DECAY_MS` env read is still there** (`heap.c:2557`). The setter now covers the tests, so the env var's only remaining audience is operators — which is precisely the ad-hoc-knob surface this project doesn't add (cf. `RAY_DIRECT_CACHE_MB`, removed in 1ae57b98). Drop the `getenv` and keep `ray_heap_set_decay_ms` as the programmatic surface; if an operator override is genuinely wanted, it should ride an existing surface (CLI flag / `.mem.*`) as its own discussion, not ship as an env var inside this fix.

2. **Stale design comment**: the block above the decay section still says "The stamp is taken at the START of a unit of work, never at the end. A stamp at the end would make the elapsed time at the immediately following boundary check zero…" — that describes the ordering 9778f14b removed, and its rationale is now backwards (the check runs before the stamp, so end-stamping is what makes the gap real). Same block still says the idle check is "a relaxed load, a getenv and a subtraction" — the getenv is gone after first resolve. `heap.h:257` also still documents the env var as the threshold's interface.

singaraiona (approved):
9e92339c closes out the review: the environment read is gone with the setter as the only surface, and both stale comments now describe the code as it is. All findings from both review rounds are addressed — approving; merging once CI is green on the updated branch.

PR #449, opened 2026-08-31
fix(join): a null equi-join key equals another null key

### LINKED ISSUE #411: update ... by: loses HAS_NULLS on the per-group gather — aggregates read null sentinels as values

protocolstardust:
## The invariant

`(anti-join [cols] X X)` must be empty for every table `X` and every key set. It was not: a null in a key column made a row unmatchable against **itself**, so a table anti-joined with itself came back holding exactly its null-keyed rows.

`left-join` has the same defect from the other side — a left row whose key is null never finds the right row that carries the same null key, so it yields a null payload instead of the match that is actually there.

## Minimal reproduction

```lisp
(set L (table [K]   (list (as 'I32 [1 2 3]))))
(set R (table [K V] (list (as 'I32 [1 2]) (as 'F64 [10.0 20.0]))))
(set J (left-join [K] L R))     ;; row K=3 gets V = null

(count (select {from: J where: (nil? V)}))  ;; 1
(count (anti-join [V] J J))                 ;; 1  — MUST be 0
(count (anti-join [K] J J))                 ;; 0  — correct, K has no nulls
```

Deterministic, and reproduced on SYM, STR, F64 and integer key columns alike — this is the shared key path, not a per-type issue.

## Root cause

Two helpers in `src/ops/join.c`, both shared by the radix and chained hash paths and by all four equi-join verbs:

- `hash_row_keys` returned `h ^ (row * 0x9E37...)` for a null key — a **per-row** hash, so two nulls never even landed in the same bucket.
- `join_keys_eq` returned `false` the moment either side was null (`/* NULL != NULL in join predicates */`).

## How it was found

`anti-join` used as a set difference — "which of these keys have I not seen yet" — over a key column that admits nulls. Every null-keyed row came back as unseen on every pass, so the caller kept re-adding rows it already held, and the id it looked up afterwards came back null. Verified against core commit 057f519; the reproduction above is independent of any particular caller.

## Why nulls should match here

This is a semantic choice, so it is worth stating why it lands on "match" rather than SQL's "unknown".

**The rest of this runtime already matches nulls on key columns.** `collection.c`'s hashset folds every null into a single null bucket, so `distinct` keeps exactly one null (`test/rfl/null/distinct.rfl`) and `in` answers true for null ∈ {…null…}; `group.c` folds nulls into one group; `atom_eq` returns 1 for two nulls. Joins were the only holdout, which made `distinct`/`group`/`anti-join` disagree about what "the same key" means.

**The join path was not even self-consistent.** The radix single-STR-key specialization compares raw `ray_str_t` descriptors via `join_str_eq_hashed`, where two zero-length (null) cells already came out **equal**. So the same single-STR-key join answered one way below `RAY_PARALLEL_THRESHOLD` and the other way above it.

**The never-match rule was never chosen for symbol keys.** It is original v2-engine code (4a3a0c7), written when "null" meant a numeric sentinel. Empty `SYM`/`STR` only became canonical nulls in bfb5b38 (`fix(core): normalize null and schema handling`, Fixes #411/#414/#415) — a commit that **never touched `src/ops/join.c`** but silently dragged symbol keys under the rule. In the same commit the stress oracle was rewritten from:

> the shadow's sum over distinct live tickers of `live_count(k) * hist_count(k)` — **including the `""` null key, which the join matches like any other (probed)**

to "Canonical empty/null keys do not join." So symbol keys *used* to join on null and stopped without anyone deciding they should. This PR restores that, and makes it uniform across every key type instead of just symbols.

**kdb+/q — where this system's surface syntax and users come from — matches nulls** in `lj`/`ij` and in `distinct`/`group`.

**The invariant is what makes anti-join useful.** "Which of these tuples have I not seen yet" is only answerable if a tuple is comparable to itself.

Nothing in the tree documented the old behaviour: the one written rule, `docs/queries/joins.md` → "Null-key semantics", is scoped to **as-of join**, and as-of join is deliberately left alone (see below).

## The change

- Nulls hash to one canonical payload per key position, so both sides land in the same bucket.
- `join_keys_eq` answers: null == null, null != non-null. The null branch is no looser than the value arms that follow it — STR pairs only with STR, GUID only with GUID.
- Both helpers are shared, so `inner-join` / `left-join` / `full-join` / `anti-join` move together across the radix and chained paths.

**As-of join is untouched.** Its NULLs-never-match rule is documented, is load-bearing (a null time must not be read as time zero), and its merge walk carries its own per-row null bitsets — it never routes through `join_keys_eq`. `docs/queries/joins.md` now states the equi-join rule alongside it, and the regression test pins that as-of behaviour did not move.

## Tests

New `test/rfl/join/null_key_equality.rfl` covers the self-anti-join invariant across I64 / F64 / SYM / STR and multi-key sets, anti-join and left-join with nulls on either side, that a null still never equals a non-null, the radix path past `RAY_PARALLEL_THRESHOLD`, and the unchanged as-of behaviour.

Two existing expectations encoded the old rule and move with it:

- `test/stress_eval.c` — the join-cardinality shadow oracle counts the `""` key again. This is an exact revert of the bfb5b38 hunk quoted above.
- `test/rfl/ops/join_branch_cov.rfl` — its two **both-sides-null** cases move: the single-key inner-join now expects 3 rather than 2, and the multi-key one expects 1 rather than 0. Everything else in that file is untouched and still passes — the one-sided-null inner-joins, the left-joins, the anti-joins, and all eight as-of assertions. That is the useful signal: only null-meets-null moved.

Full suite green before and after: `3713 of 3714 passed (1 skipped, 0 failed)` before, `3714 of 3715 passed (1 skipped, 0 failed)` after (+1 is the new file).

> Note on the local run: this machine is macOS 26.6.2 with Apple clang 17, where **any** `-fsanitize=address` binary hangs in `__asan::InitializeShadowMemory` at startup (a bare hello-world reproduces it). The suite was therefore built and run with the sanitizers disabled — same objects, same tests, `DEBUG_CFLAGS`/`DEBUG_LDFLAGS` overridden. CI's ASan run is the one to trust for sanitizer coverage.

### DIFF
--- a/docs/docs/queries/joins.md
+++ b/docs/docs/queries/joins.md
@@ -126,6 +126,22 @@ The `anti-join` function keeps rows from the left table whose key does not appea
 ;  3 MSFT
 ```
 
+## Null-key semantics (equi-joins)
+
+`inner-join`, `left-join`, `full-join` and `anti-join` treat **a null key as equal to another null key**, and never equal to a non-null one:
+
+```lisp
+(set left  (table [k v] (list (as 'SYM ["a" ""]) [1 2])))
+(set right (table [k w] (list (as 'SYM ["a" ""]) [10 20])))
+
+(at (left-join [k] left right) 'w)  ; [10 20]  — the null key finds its match
+(count (anti-join [k] left left))   ; 0        — X anti-joined with itself is empty
+```
+
+This is the rule the rest of the runtime already applies to key columns: `distinct` keeps exactly one null, group-by folds nulls into a single group, and `in` matches a null against a null. It differs from SQL, where `NULL = NULL` is unknown and a self-anti-join can return rows. The invariant that follows — `(anti-join [cols] X X)` is empty for every `X` and every key set — is what makes anti-join usable for "which of these tuples have I not seen yet".
+
+As-of join is the exception — its null keys never match. See its own [Null-key semantics](#null-key-semantics) section below.
+
 ## As-of Join
 
 The `asof-join` function is designed for time-series data. For each row in the left table, it finds the most recent matching row in the right table where the time column is less than or equal to the left's time value. The last column in the key list is the temporal column; any preceding keys are exact-match equality keys.
--- a/src/ops/join.c
+++ b/src/ops/join.c
@@ -102,16 +102,26 @@ static inline bool join_str_eq_hashed(const ray_str_t* a, const char* pool_a,
     return memcmp(pa, pb, a->len) == 0;
 }
 
+/* Hash payload for a null key cell.  One canonical value per key POSITION
+ * (the ray_hash_combine chain below still mixes in k), so every null in a
+ * given key column lands in the same bucket on both sides of the join —
+ * the precondition for join_keys_eq's null == null.  Any collision with a
+ * real key's hash is resolved by join_keys_eq, as for any other hash. */
+#define JOIN_NULL_KEY_HASH INT64_C(0x5B7A6D3F2E1C0A94)
+
 static uint64_t hash_row_keys(ray_t** key_vecs, uint32_t n_keys, int64_t row) {
     uint64_t h = 0;
     for (uint32_t k = 0; k < n_keys; k++) {
         ray_t* col = key_vecs[k];
         if (!col) continue;
-        /* NULL key — produce unique hash that won't match any other row */
-        if (ray_vec_is_null(col, row))
-            return h ^ ((uint64_t)row * 0x9E3779B97F4A7C15ULL);
         uint64_t kh;
-        if (col->type == RAY_STR) {
+        if (ray_vec_is_null(col, row)) {
+            /* null == null: hash every null cell alike instead of giving the
+             * row a private hash.  A per-row hash made a null key unmatchable
+             * even against itself, so `anti-join [c] X X` came back non-empty
+             * and a null-keyed left-join row missed its real match. */
+            kh = ray_hash_i64(JOIN_NULL_KEY_HASH);
+        } else if (col->type == RAY_STR) {
             const char* pool = NULL;
             const ray_str_t* str = join_str_cell(col, row, &pool);
             kh = ray_str_is_inline(str)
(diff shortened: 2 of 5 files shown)


singaraiona (changes requested):
The core change is correct on its mainline paths — every chained and radix call site routes through the patched `hash_row_keys`/`join_keys_eq` pair, the i64 bitmap fast paths pass the sentinel through, the single-STR radix specialization agrees with the new semantics today, and the stress-oracle change is right. Two things need attention before this merges:

1. **The null==null arm is looser than the value arms** (`join.c:605`): it gates only STR and GUID pairing, so nulls of any *other* mixed type pair compare equal — F64 `0N` vs I64 `0N`, SYM vs I64, DATE vs F64. Join setup (`query.c:14633`) validates only STR-vs-STR, so a schema-drifted feed reaches it: an inner join then emits phantom rows exactly and only on nulls, while no non-null cross-type pair can match. Require the same type class in the null arm (or add a setup-time key-compatibility check).

2. **Canonical null hash creates a silent radix abort** (`join.c:815`): all nulls in a key column now share one hash, so a build side with >512 nulls (`RADIX_DUP_RUN_MAX`) trips the pathological-duplication abort and re-runs the whole join on the chained path — full hash/partition/build work discarded, every execution, for what's a routine 1%-nulls key at ≥64k rows. Pre-PR nulls had unique per-row hashes and never formed a run. Exempt/pre-count nulls (they're mutually equal, a null probe needs no HT walk) or gate radix eligibility on build-side null count — and at minimum benchmark a nullable-key join at scale before this lands.

Non-blocking, worth folding in:
- `docs/joins.md:141` — "anti-join [cols] X X is empty for every X" is conditional on HAS_NULLS flag hygiene: an unflagged NaN F64 key fails `is_null` on both sides and then NaN != NaN in the value arm, so the row matches nothing. Soften the claim; a HAS_NULLS differential join test would pin it (same class as 3ebf8f23/0fb4d012).
- The single-STR radix path achieves null==null only by representation coincidence (len-0 descriptor short-circuit), and the new test's tables are too small to exercise it — add a >65536-row STR variant of the NKBIG block so the big-table path can't silently diverge.
- `JOIN_NULL_KEY_HASH` is a pre-image fed through `ray_hash_i64`, not the hash itself, and it's the runtime's third private null-folding scheme (group.c's `ght_hash_null_words`, collection.c's null bucket). Name it as the pre-image or hoist a shared helper.
- `join_branch_cov.rfl:83` — the anti-join comments still state the deleted nulls-never-match semantics; the assertions pass only because the opposite side has no null.

singaraiona (approved):
Re-reviewed the rework: the null arm's STR/GUID type gates make it exactly as strict as the value arms (integer widths compare by value, so cross-width null==null is the consistent choice), the single-STR radix/chained inconsistency is fixed, the SYM cross-domain translation matches the hashset precedent, and the new test file pins the invariant across types, both hash paths, and the as-of exception. The remaining item from round one — the >512-null radix dup-run abort — is a perf follow-up, not a correctness blocker (the pre-#449 radix speed on those joins was the speed of a wrong answer); filed as #458. Approving.

PR #451, opened 2026-08-31
fix(mem): size the scratch arena window from the backing block's real capacity

ser-vasilich:
The scratch arena computed the end of its window as `blk + BSIZEOF(blk->order)`. `order` describes the size of a buddy block; a request at or above the pool order is not one. Those are served by `heap_alloc_direct`, which maps the exact page-rounded size and tags the header with the `RAY_ORDER_DIRECT` sentinel instead of an order.

Read as an order, that sentinel claims a block of 2^39 bytes. Every push after the one that crossed the threshold is then bump-allocated past the end of the mapping, and the caller's first write goes straight into unmapped address space.

Nothing about the arena's users is special here — any arena that crosses the threshold corrupts memory the same way. Sorting by a SYM column reaches it first because ranking the vocabulary pushes four buffers sized by the number of distinct symbols, so a store with millions of them crosses on one of them. A small vocabulary keeps every push inside a buddy block, which is why this stayed invisible.

`ray_block_data_bytes` already answers the capacity question for both kinds of block. The window is now anchored at the data pointer and sized from it, so the arithmetic no longer depends on the header size either. For buddy blocks the result is byte-identical, so nothing changes on the path this normally takes.

## Verification

`heap/scratch_arena_direct` allocates one backing past the pool order, asserts the window stops at the block's real capacity, and checks that the following push still lands inside it. On the parent commit it fails:

```
test/test_heap.c:312: (a.end <= cap_end): expected true
```

End to end, on a splayed store whose SYM vocabulary runs into the millions: sorting by that column dumps core on the parent and returns its rows after the fix. The failure is intermittent on a release build — the wild store has to reach an unmapped page before anything else notices — and immediate under ASan, which names the store in `build_enum_rank`. Across five runs on the parent it faulted once; with the fix, five out of five completed.

The end-to-end shape needs a vocabulary too large for the suite, so the committed test pins the invariant rather than the symptom.

### DIFF
--- a/src/mem/heap.c
+++ b/src/mem/heap.c
@@ -2689,7 +2689,15 @@ void* ray_scratch_arena_push(ray_scratch_arena_t* a, size_t nbytes) {
     if (!blk) return NULL;
     a->backing[a->n_backing++] = blk;
     a->ptr = (char*)ray_data(blk);
-    a->end = (char*)blk + BSIZEOF(blk->order);
+    /* The window ends at the block's real capacity.  `order` describes the
+     * block size only for buddy blocks: a request at or above
+     * RAY_HEAP_POOL_ORDER comes back as a DIRECT block, mapped at its exact
+     * page-rounded size and tagged with the RAY_ORDER_DIRECT sentinel.
+     * Reading that sentinel as an order claims a 2^39-byte block, and every
+     * push after the one that triggered it is then bump-allocated past the
+     * end of the mapping — a write straight into unmapped address space.
+     * ray_block_data_bytes covers both block kinds. */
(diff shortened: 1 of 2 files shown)


singaraiona (approved):
Verified: for buddy blocks `ray_block_data_bytes` reproduces the old `BSIZEOF(order)` window bit-for-bit, and for DIRECT blocks it ends the window at the real mapping end instead of 2^39; the fresh-block bump is bounded by the page-rounded alloc size, and the 33MB test lands on the direct path (order 26 ≥ pool order 25). Merging.

PR #456, opened 2026-09-01
fix(query): handle guid membership predicates

belowzeroff:

A user filters a GUID column with `in` or `not-in`:

```rfl
(select {from: trades where: (not-in guid_col target_guids)})
```

Before, GUID membership could silently produce incorrect matches. In particular, a GUID atom used as the subject could return a scalar with the wrong `not-in` result, and null GUID rows could pass `not-in` because the all-zero GUID matched the fallback hashset and was inverted.

After, GUID membership compares the complete 16-byte values, returns a row-aligned BOOL vector from the query execution path, and null GUID rows remain excluded from both `in` and `not-in` predicates.

## Root Cause

The typed `in` worker only supports numeric and SYM readers, so GUID operands must use the GUID-aware fallback. The fallback previously assumed vector-shaped results and applied negation without preserving the typed null semantics used by the worker. Parted operands also exposed segment pointers to generic membership code instead of flat GUID cells.

## Fix

- Route GUID operands through the GUID-aware fallback.
- Normalize scalar fallback results to a one-row BOOL vector for `exec_in`.
- Apply `not-in` before masking canonical all-zero GUID null cells, so null rows stay false.
- Flatten parted operands through a shared path before membership checks.
- Correct the parted-membership comment to describe only the operation that actually shares this flattening behavior.

## Tests

- `RAYFORCE_CORES=2 ./rayforce.test -f rfl/query/guid_like_predicates`
- `./rayforce.test -f lang/eval/select_where_in_guid_nulls`
- `make debug`

The two targeted tests pass. A full local `make test` reaches the end but currently reports unrelated IPC/stress/system failures in the local test environment; the GUID regression tests themselves pass.

### DIFF
--- a/src/ops/collection.c
+++ b/src/ops/collection.c
@@ -1224,6 +1224,25 @@ ray_t* ray_distinct_fn(ray_t* x) {
 ray_t* ray_in_fn(ray_t* val, ray_t* vec) {
     if (ray_is_lazy(val)) val = ray_lazy_materialize(val);
     if (ray_is_lazy(vec)) vec = ray_lazy_materialize(vec);
+    /* Parted columns store segment pointers where vector paths expect flat
+     * cells: reading a parted GUID column through ray_data() yields segment
+     * pointers, not 16-byte guids, so membership silently matched nothing.
+     * Flatten first, then run the normal path.  This mirrors distinct/except/
+     * sect, which already flatten via parted_to_flat_vec. */
+    if (val && RAY_IS_PARTED(val->type)) {
+        ray_t* fv = parted_to_flat_vec(val);
+        if (!fv || RAY_IS_ERR(fv)) return fv ? fv : ray_error("oom", NULL);
+        ray_t* r = ray_in_fn(fv, vec);
+        ray_release(fv);
+        return r;
+    }
+    if (vec && RAY_IS_PARTED(vec->type)) {
+        ray_t* fs = parted_to_flat_vec(vec);
+        if (!fs || RAY_IS_ERR(fs)) return fs ? fs : ray_error("oom", NULL);
+        ray_t* r = ray_in_fn(val, fs);
+        ray_release(fs);
+        return r;
+    }
     /* STR in STR: for each char of val, check membership in vec string */
     if (ray_is_atom(val) && (-val->type) == RAY_STR && ray_is_atom(vec) && (-vec->type) == RAY_STR) {
         const char* vp = ray_str_ptr(val);
--- a/src/ops/collection.c
+++ b/src/ops/collection.c
@@ -1225,22 +1225,27 @@ ray_t* ray_in_fn(ray_t* val, ray_t* vec) {
     if (ray_is_lazy(val)) val = ray_lazy_materialize(val);
     if (ray_is_lazy(vec)) vec = ray_lazy_materialize(vec);
     /* Parted columns store segment pointers where vector paths expect flat
-     * cells: reading a parted GUID column through ray_data() yields segment
-     * pointers, not 16-byte guids, so membership silently matched nothing.
-     * Flatten first, then run the normal path.  This mirrors distinct/except/
-     * sect, which already flatten via parted_to_flat_vec. */
+     * cells.  Flatten both operands before membership; distinct uses the same
+     * helper, while except/sect still reject parted input. */
+    ray_t* flat_val = NULL;
+    ray_t* flat_vec = NULL;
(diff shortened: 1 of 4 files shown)


singaraiona (changes requested):
The routing itself is sound (no recursion loop — `ray_in_vec_exec` returns NULL for guid and `ray_in_fn` lands in the guid-aware hashset), and flattening parted operands through the existing `parted_to_flat_vec` mirrors distinct/except/sect correctly. But two correctness gaps need fixing before merge:

1. **`not-in` on a guid atom subject returns the un-negated answer** (`exec.c`, `exec_in_guid_fallback`): for an atom `col`, `ray_in_fn` returns an atom bool (type `-RAY_BOOL`), so `if (negate && r->type == RAY_BOOL)` never fires — `(not (in g gset))` with a scalar guid silently returns the `in` result. It also violates `exec_in`'s contract of returning a BOOL *vector* of `col_len` (1 for atoms), which downstream vector consumers won't expect. Fix: normalize the atom result to a length-1 BOOL vector and negate that.

2. **Null semantics diverge from the typed kernel**: `exec_in_worker`'s paths enforce "null rows emit 0 regardless of negate" (the comment in `exec_in` about the old memset bypass exists precisely for this), but `ray_in_fn`'s hashset path matches null==null as TRUE — and the blind bit-invert then turns a null guid cell into TRUE under `not-in` even when no null is in the set. A guid column with nulls now answers membership differently from every other column type. Fix: mask null rows to 0 in the fallback after `ray_in_fn` (before the negate), using the column's null state.

Neither is exotic: a null guid cell is just an all-zero guid, and scalar-guid predicates are ordinary. `guid_like_predicates.rfl` has no null-guid or atom-subject `in`/`not-in` case — both fixes should come with one.

Non-blocking cleanups worth folding in:
- `in_is_guid` re-derives the exact atom/parted normalization `in_build_worker_ctx` already computed, 250 lines away, just to split "unsupported because STR" from "unsupported because GUID". Split the status (`IN_CTX_UNSUPPORTED_STR`/`_GUID`) or add an out-param instead — two copies of the classification will drift.
- The `val`-parted and `vec`-parted flatten blocks in `ray_in_fn` are copy-paste with one variable swapped; a single flatten-both-sides pass halves it.
- The new parted branches dereference `->type` after `ray_lazy_materialize` without an `RAY_IS_ERR` check, unlike the mature branches in the same function.

singaraiona:
One addition to the review: the new comment on the parted-flatten block says it "mirrors distinct/except/sect, which already flatten via parted_to_flat_vec" — only `ray_distinct_fn` actually does; `except`/`sect`/`union` never touch `parted_to_flat_vec` and error on parted input via the boxed-list fallback (`find` too). Fix the comment to name only distinct, and consider a follow-up for the others — parted handling is being bolted on per-op, and this PR's comment would mislead the next reader into thinking they're covered.

singaraiona:
Two performance notes from the deeper pass, for the same revision (not blockers, but worth addressing while in here since both are specific to the new routing):

- **Guid IN predicates lose the fused selection path**: since `in_build_worker_ctx` bails to `IN_CTX_UNSUPPORTED` for guid, `exec_in_to_selection` returns NULL and every `where (in g ...)` on a guid column pays a full n-byte BOOL materialization plus a second scan to rebuild the row selection — a permanent ~2x memory-traffic tax that non-guid IN filters on the same shape don't pay.
- **The fallback probe is single-threaded**: `ray_in_fn`'s hashset loop has no `ray_pool_dispatch`, so a large guid IN filter runs on one core where the typed kernel parallelizes past `RAY_PARALLEL_THRESHOLD` — a silent wall-clock cliff proportional to core count.

Both would be resolved more durably by teaching the typed kernel a 16-byte key mode (or giving the fallback a dispatched probe) rather than growing the `ray_in_fn` detour — worth considering when fixing the two correctness items from the review.

singaraiona (approved):
Re-reviewed f36cf844: both blockers are properly fixed — the scalar result is normalized to a length-1 BOOL vector before the negate (so `not-in` on a guid atom is correct and the vector contract holds), and null rows are masked to 0 after inversion for both IN and NOT-IN, with payload-truth masking for guid columns so flag-less legacy vectors are covered too. The parted flatten moving into the fallback also keeps `ray_in_fn` un-detoured. Approving; will merge after the branch updates against dev. The perf notes (fused-selection bypass, single-threaded probe) stand as follow-up material for a 16-byte key mode in the typed kernel.

Issue #363, opened 2026-08-03
`-U` read-only mode is bypassable — a client gets shell + arbitrary file access via a lambda


birrgrrim:
### Summary

In read-only IPC mode (`-U`), the restricted-builtin check only fires when a
restricted verb is the **head of the form the client sent**. Wrapping the
same call in a lambda bypasses it entirely, giving a connected client shell
execution, arbitrary file read/write, and mutation of server state.

This is not embedding-specific — it reproduces on a stock `rayforce -U`
server.

### Environment

- Version: **v2.5.7** (`ceac936a`)
- OS: macOS (kqueue); the gate is platform-independent (`src/lang/eval.c`)

### What the docs promise

From `docs/docs/guides/ipc.md`:

> In read-only mode, clients can run queries and read data, but mutating
> operations (`set`, `insert`, `upsert`, `update`, file writes, system
> commands) are blocked.

### Reproduction

Server:

```bash
rayforce -p 7799 -U secretpass
```

Client:

```clj
(set h (.ipc.open "127.0.0.1:7799:u:secretpass"))

;; direct call — correctly refused
(.ipc.send h "(.sys.exec \"touch /tmp/pwn\")")
;; => error: access

;; same call inside a lambda — RUNS
(.ipc.send h "((fn [x] (.sys.exec \"touch /tmp/pwn\")) 0)")
;; => 0        (/tmp/pwn created)

;; arbitrary file write — 5 bytes land on disk
(.ipc.send h "((fn [x] (write \"/tmp/pwn\" \"owned\")) 0)")
;; => 0

;; arbitrary file read
(.ipc.send h "(count ((fn [x] (read \"/etc/hosts\")) 0))")
;; => 213

;; mutate a server global, and it persists
(.ipc.send h "((fn [x] (set 'injected 1)) 0)")
;; => 1
(.ipc.send h "injected")
;; => 1
```

Every restricted builtin we tried is reachable this way: `set`, `insert`,
`upsert`, `.db.splayed.set`, `read`, `write`, `load`, `exit`, `.sys.exec`.

Note the current gate *does* catch `(eval (list 'set …))` (returns `access`)
— so obvious indirection is handled, but an ordinary lambda body is not.

### Where it is

`src/lang/eval.c`, `fn_is_restricted` (line ~123). The `access` refusals fire
at the direct-dispatch call sites (around lines 1358, 1386, 2491–2522,
3693–3803). When a restricted builtin is invoked from **inside a lambda
[...]

PR #464, opened 2026-09-03
fix(ipc): restrict server-state mutators under -U restricted IPC

belowzeroff:

An operator runs a server in restricted mode so untrusted clients get read-only access. The server keeps live vectors, graph/HNSW handles, and Datalog program handles in globals.

Before this fix, a restricted client could send calls such as `(alter 'g set 0 999)`, `(.graph.free g)`, `(hnsw-free h)`, `(dl-add-edb p ...)`, `(dl-stratify p)`, `(dl-eval p)`, `(dl-free p)`, `(rule ...)`, or `(.log.write ...)`. The server accepted these calls, so the client could overwrite global data, destroy live handles, mutate or recompute server-held Datalog state, register a global rule, or persist WAL work.

After this fix, the same calls return an `access` error during restricted evaluation. The server-side values and handles remain unchanged; normal local evaluation, startup scripts, and WAL replay continue to work.

## Root cause

A `-U` restricted IPC client's evaluations run with `__VM->restricted` set, and `fn_is_restricted` blocks builtins carrying `RAY_FN_RESTRICTED`. Nine mutating or destructive registrations were missing that flag:

| builtin | protected state |
|---|---|
| `alter` | global vectors/lists |
| `rule` | global Datalog rules |
| `dl-add-edb` | Datalog program relations |
| `dl-stratify` | Datalog program strata |
| `dl-eval` | Datalog program IDB results |
| `dl-free` | Datalog program handles |
| `.graph.free` | graph handles |
| `hnsw-free` | HNSW handles |
| `.log.write` | WAL entries replayed outside restricted mode |

## Fix

Add `RAY_FN_RESTRICTED` to all nine registrations. Special forms retain their existing `RAY_FN_SPECIAL_FORM` attribute. The gate only fires while `__VM->restricted` is set.

## Tests

`lang/eval/restricted_server_state_ops` now checks all nine verbs return an `access` error under restricted evaluation, while normal-mode behavior remains available. Full `make test`: 3729/3730 pass (1 skipped).

### DIFF
--- a/src/lang/eval.c
+++ b/src/lang/eval.c
@@ -3335,7 +3335,7 @@ static void ray_register_builtins(void) {
     register_unary("guid",       RAY_FN_NONE, ray_guid_fn);
 
     /* In-place mutation */
-    register_vary("alter",       RAY_FN_SPECIAL_FORM, ray_alter_fn);
+    register_vary("alter",       RAY_FN_SPECIAL_FORM | RAY_FN_RESTRICTED, ray_alter_fn);
 
     /* Pattern matching */
     register_binary("like",      RAY_FN_NONE, ray_like_fn);
@@ -3490,8 +3490,8 @@ static void ray_register_builtins(void) {
     /* Programmatic Datalog API */
     register_vary("dl-program",    RAY_FN_NONE, ray_dl_program_fn);
     register_vary("dl-add-edb",    RAY_FN_RESTRICTED, ray_dl_add_edb_fn);
-    register_unary("dl-stratify",  RAY_FN_NONE, ray_dl_stratify_fn);
-    register_unary("dl-eval",      RAY_FN_NONE, ray_dl_eval_fn);
+    register_unary("dl-stratify",  RAY_FN_RESTRICTED, ray_dl_stratify_fn);
+    register_unary("dl-eval",      RAY_FN_RESTRICTED, ray_dl_eval_fn);
     register_binary("dl-query",    RAY_FN_NONE, ray_dl_query_fn);
     register_binary("dl-provenance", RAY_FN_NONE, ray_dl_provenance_fn);
     register_unary("dl-free",      RAY_FN_RESTRICTED, ray_dl_free_fn);
--- a/src/lang/eval.c
+++ b/src/lang/eval.c
@@ -3417,7 +3417,7 @@ static void ray_register_builtins(void) {
      * startup; these builtins expose the same machinery to Rayfall code
      * for manual control (open from a script, snapshot on demand, etc). */
     register_vary(".log.open",     RAY_FN_RESTRICTED, ray_log_open_fn);
-    register_unary(".log.write",   RAY_FN_NONE,       ray_log_write_fn);
+    register_unary(".log.write",   RAY_FN_RESTRICTED, ray_log_write_fn);
     register_unary(".log.replay",  RAY_FN_RESTRICTED, ray_log_replay_fn);
     register_unary(".log.validate",RAY_FN_NONE,       ray_log_validate_fn);
     register_vary(".log.roll",     RAY_FN_RESTRICTED, ray_log_roll_fn);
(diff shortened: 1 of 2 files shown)


src/lang/eval.c
```diff
@@ -3484,17 +3484,17 @@ static void ray_register_builtins(void) {
...
-    register_vary("rule",         RAY_FN_SPECIAL_FORM, ray_rule_fn);
+    register_vary("rule",         RAY_FN_SPECIAL_FORM | RAY_FN_RESTRICTED, ray_rule_fn);
     register_vary("query",        RAY_FN_SPECIAL_FORM, ray_query_fn);
 
     /* Programmatic Datalog API */
     register_vary("dl-program",    RAY_FN_NONE, ray_dl_program_fn);
-    register_vary("dl-add-edb",    RAY_FN_NONE, ray_dl_add_edb_fn);
+    register_vary("dl-add-edb",    RAY_FN_RESTRICTED, ray_dl_add_edb_fn);
     register_unary("dl-stratify",  RAY_FN_NONE, ray_dl_stratify_fn);
     register_unary("dl-eval",      RAY_FN_NONE, ray_dl_eval_fn);
```
singaraiona: `dl-stratify` and `dl-eval` remain `RAY_FN_NONE`, so a restricted client can still mutate a server-held Datalog program handle even though `dl-add-edb` on the same handle is now denied.

`ray_dl_stratify_fn` / `ray_dl_eval_fn` (src/ops/datalog.c:4466, 4474) call `dl_stratify` / `dl_eval`, which write into the program in place: strata are rewritten and IDB relations recomputed and overwritten. That is the same "mutate a Datalog program" escalation the PR table lists for `dl-add-edb`. A client can clobber IDB tables the server built, or burn CPU on an arbitrary fixpoint run. Reproduced live:

```
server: (set p (dl-program))
client: (dl-stratify p) -> true
client: (dl-eval p)     -> true
```

Suggest flagging both `RAY_FN_RESTRICTED` (`dl-program` / `dl-query` / `dl-provenance` can stay open) and adding them to the new test.

singaraiona (commented):
Reviewed the diff and verified the gates on a live restricted server (`./rayforce -U seekrit -p 19987`, client authenticated via `.ipc.open`). The six flag additions and the test are correct. Two verbs the PR does not touch still allow the same class of escalation, so the stated goal ("restrict server-state mutators under -U") is not yet met.

**1. `alter` is still `RAY_FN_SPECIAL_FORM` without `RAY_FN_RESTRICTED`** (`src/lang/eval.c:3338`, outside this diff so noted here rather than inline).

`ray_alter_fn` (src/ops/tblop.c:1238) resolves the target with `ray_env_get` and commits with `ray_env_set` (src/lang/env.c:697), which has no restricted check; only the `set` builtin and `OP_STOREGLOBAL` are gated. A restricted client can therefore rewrite any server global in place even though `set` is denied. Reproduced live:

```
server: (set g [10 20 30])
client: (set g 0)          -> denied
client: (alter 'g set 0 999) -> 'g
server: g                  -> [999 20 30]
```

Suggest adding `| RAY_FN_RESTRICTED` to the `alter` registration alongside `update`/`insert`/`upsert`/`del`, and extending `test_eval_restricted_server_state_ops` with `(alter 'x set 0 1)`.

Since this is the third round of per-verb patching (#363, this PR, now `alter`), a registration-time audit test asserting that every special form which can reach `ray_env_set` carries the flag would stop the next gap.

**2. `dl-stratify` / `dl-eval`** — see inline comment.

singaraiona (approved):
Verified 0a442487 on a fresh build: `make test` passes (3733/3733), and against a live `-U` server an authenticated restricted client now gets `error: access` for `(alter 'g set 0 999)`, `(dl-stratify p)` and `(dl-eval p)`, with the server global left unchanged and `dl-query` still open. Both findings resolved.

PR #469, opened 2026-09-04
fix(query): count a null as one distinct value in the grouped parallel kernel

ser-vasilich:
`select {d: (count (distinct v)) by: k}` answers differently depending on how many cores the process was given. The parallel per-group kernel skipped null cells; the serial fallback beside it, the ungrouped builtin, and the rule already written down in `test/rfl/null/distinct.rfl` all keep at most one null and count it.

```lisp
(set t (table [k v] (list [0 1 2 3 0 1 2 3 0 1 2 3]
                          [10 0N 20 30 0N 20 0N 30 10 20 30 0N])))
(print (at (select {d: (count (distinct v)) by: k from: t}) 'd))
```

|  | before | after |
|---|---|---|
| `-c 1` | `[2 2 3 2]` | `[2 2 3 2]` |
| `-c 8` | `[1 1 2 1]` | `[2 2 3 2]` |

The counts are checkable by hand: group 0 takes rows 0, 4, 8 → `10, 0N, 10` → two distinct values; group 2 takes rows 2, 6, 10 → `20, 0N, 30` → three. One null lost per group.

The kernel is chosen when the pool has at least two workers and the query has at least four groups, which is why the smaller shape shows nothing and why `-c 1` looks correct. The default run is multi-core, so the wrong answer was the default one — returned silently, with no error.

## The fix

The kernel already had the shape this needed. Its hash table encodes an empty slot as `0`, so the literal `0` cannot be stored and rides a side flag instead. A null has the same problem for the same reason, so it now rides its own flag beside it: seen once, counted once, however many rows carry it.

The loops taken when a column has no nulls are untouched. On a null-bearing column the new work replaces a `continue`, so nothing is added to the hot path.

This also settles a split between types. A SYM null is the empty symbol, id `0`, so it was already being counted — by accident, through the flag kept for the literal zero — while integer, float and temporal nulls were dropped. Both follow one rule now.

The comment above the kernel claimed that null-bearing columns fall through to the serial path. No such gate exists; the kernel handles them itself, which is how the drift arose. Corrected to describe what the code does.

## Why five green tests did not catch it

`count_distinct_fused.rfl`, `ungrouped_count_distinct.rfl`, `agg/count_distinct.rfl`, `count_distinct_extras.rfl` and `regress/fuzz_count_distinct_f64.rfl` all pass, and none of them contains a null. The rule was pinned only against the ungrouped builtin, while the grouped kernel is a separate implementation of it.

So the regression cases go into `test/rfl/null/distinct.rfl`, next to the rule they enforce, and are built to reach the kernel — four groups each, and the suite runs with `TEST_CORES=2`. They cover the integer arm, a group that is entirely null, the F64 arm, and the SYM arm; the last is pinned because it passes through the zero flag rather than the null one, so the two need holding together or they drift apart again.

On the parent commit the first case fails with `[1 1 2 1]` against `[2 2 3 2]`.

Full suite: 3733/3733 (ASan + fatal UBSan).


src/ops/query.c
```diff
@@ -3473,7 +3485,7 @@ static void cdpg_buf_par_fn(void* vctx, uint32_t worker_id,
             if (has_nulls) {
                 for (int64_t i = 0; i < cnt; i++) {
                     int64_t v = d[idxs[i]];
-                    if (v == NULL_I64) continue;
+                    if (v == NULL_I64) { CDPG_BUF_NULL(); continue; }
```
singaraiona: **Simplification:** the `saw_null` flag / `CDPG_BUF_NULL` is only needed in the F64 arm. For esz 8/4/2 the null sentinels (`NULL_I64` / `NULL_I32` / `NULL_I16`) are INT_MIN values: nonzero and unique after widening to `int64_t`, so the plain `!has_nulls` loop (`CDPG_BUF_INSERT((int64_t)d[idxs[i]])`) already counts a null exactly once as an ordinary key. That is what `cd_seq_count` in group.c does.

Suggest running the plain insert loop unconditionally for the three integer widths and keeping the null special case only in the F64 arm (where NaN payloads can differ, so canonicalising to one payload or keeping the flag there is required). That deletes `saw_null`, `CDPG_BUF_NULL`, its `#undef`, and the three has-nulls loop bodies, which currently also carry a copy-paste inconsistency (esz 8/4 split into two loops, esz 2 folds the check inline). The rationale comment is also duplicated between the struct header and the macro.

Checked behaviour-preserving, including SYM W16/W32 columns read as signed where code `0x8000` / `0x80000000` aliases the sentinel: both forms count it once.

test/rfl/null/distinct.rfl
singaraiona, line 1: **Correctness (blocking):** these cases only exercise the `<= 50000` group route. The same query is routed to `ray_count_distinct_per_group` in group.c when `n_groups > 50000` (`src/ops/query.c` around line 10940), and that kernel still drops nulls: `if (cdpg_is_null(base, r, in_type, esz)) continue;` in the serial has-nulls loop (`src/ops/group.c:2091`) and the same skip in the partitioned path (`src/ops/group.c:1637`). So after this PR the answer depends on group cardinality instead of core count, and the "one rule on every path" claim does not hold yet.

Verified on this branch (release and debug, `RAYFORCE_CORES=1/8` and `-c 1`):

```lisp
(set t (table [k v] (list (% (til 120002) G) (at [10 0N] (% (til 120002) 2)))))
(distinct (at (select {d: (count (distinct v)) by: k from: t}) 'd))
```

| G | route | result |
|---|---|---|
| 40001 | query.c (fixed) | `[2]` |
| 60001 | group.c | `[1]` |

Every group holds `{10, 0N}` in both. A column that is entirely `0N` gives `0` per group at G=60001 versus `1` at G=40001.

Suggest applying the same rule in both group.c loops and adding a case here with more than 50000 groups so the two kernels are pinned to the same answer.

singaraiona (changes requested):
Two findings, both verified on the branch build. Details in the inline comments.

1. **Blocking:** the fix covers only the query.c kernel. Queries with more than 50000 groups route to `ray_count_distinct_per_group` in group.c, which still skips nulls in both its serial and partitioned loops, so the result now diverges by group cardinality. The new tests only hit the small-group route.
2. **Simplification:** the `saw_null` / `CDPG_BUF_NULL` machinery is redundant for the integer widths, since the INT_MIN sentinels already hash as one ordinary key. Only the F64 arm needs a null special case.

singaraiona:
Reviewed and built the branch. The change is correct for the kernel it touches, but it does not yet make the grouped answer path-independent: it replaces a core-count dependence with a group-count dependence.

**1. Blocker: the other grouped kernel still drops nulls.**

`query.c:10944` routes a direct-column `(count (distinct v)) by: k` to `ray_count_distinct_per_group` in `group.c` whenever `n_groups > 50000`. That kernel still skips nulls at three sites: the serial has-nulls loop (`group.c:2091`, `if (cdpg_is_null(...)) continue;`) and both arms of its parallel variant (`cdpg_hist_fn` at ~1637 and `cdpg_scat_fn` at ~1660). Verified on this branch, each group holding {value, 0N}:

| groups | count per group |
|---|---|
| 4 | 2 |
| 50000 | 2 |
| 50001 | 1 |
| ungrouped `(count (distinct [1 0N]))` | 2 |

Identical at `RAYFORCE_CORES=2` and `=0`. Fix the three skip sites in `group.c` the same way, and add a >50000-group case to `null/distinct.rfl` so both routes are pinned.

**2. Blocker: an existing test pins the old convention.**

`test/rfl/agg/count_distinct_extras.rfl:108-146` states "CONVENTION: the per-group count(distinct) has_nulls kernel SKIPS nulls" and asserts group 0 = {0Nl, 0} -> 1 and `(sum n) = 59996` for I64 and F64. It passes today only because `Ggrp` has 51000 groups and lands on the unfixed kernel. Once (1) is fixed these assertions flip to 2 / 59999; they need updating in the same PR.

**3. Simplification: the side flag is not needed.**

`NULL_I16/I32/I64` are `INT*_MIN`, non-zero after sign-extension to int64, so they are already legal keys in this kernel's 0-means-empty table. That is exactly how the serial `cd_seq_count` (`group.c:922-975`) counts them: it inserts the sentinel as an ordinary key. Calling `CDPG_BUF_INSERT(v)` for nulls yields the identical count and lets you delete `CDPG_BUF_NULL`, `saw_null`, `ctx->has_nulls`, and the duplicated has-nulls/no-nulls loop pairs for esz 8 and 4. Only F64 needs a branch, and the established idiom is the one used at `group.c:794/893/953/1268`: `if (fv != fv) fv = (double)NAN; else fv = clear_neg_zero(fv);` then hash. The comment "a null's payload is not a usable key" is not true for any integer width.

**4. Note, unconfirmed.** The F64 arm folds NaN only when `has_nulls` is set, and that is read from `src->attrs` directly rather than via `payload_may_have_nulls` (same file, ~11430), which exists because slices do not carry `HAS_NULLS`. If an F64 column reaches the kernel without the flag, two NaN encodings (0x7FF8.. from 0Nf, 0xFFF8.. from arithmetic) count as 2 while the serial path counts 1. None of us could manufacture such a column from Rayfall, so this is pre-existing and informational; adopting the unconditional canonicalisation in (3) removes the question.

**5. Stale comments.** `test/rfl/ops/hll_coverage.rfl:228-249` describes the deleted `continue` lines and says nulls are dropped; those queries now route through the fixed kernel and count 11 not 10, hidden by the [9,12] tolerance band.

No efficiency concern: release-build disassembly of `cdpg_buf_par_fn` shows the no-null loops instruction-for-instruction unchanged and the null path compiles to three branchless ops.

singaraiona (commented):
**Verdict: correct, recommend merge. Comments below are cleanups, none blocking.**

Checked the core claim by hand and empirically. The integer sentinels are `INT*_MIN`, nonzero and unique after widening, so they hash as ordinary keys; F64 folds every NaN to one canonical bit pattern in `cdpg_read`, the serial arm and the new query.c arm alike; the SYM null (id 0) is caught by `saw_zero` in query.c and is a legal value in group.c, whose slots are keyed on `gid+1`. The ungrouped builtin never skipped nulls, so `-c 1` and the serial fallback agree with the fixed kernels.

One gap worth closing in this PR: **no test reaches the partitioned kernel** whose null-skip lines you deleted (`cdpg_hist_fn` / `cdpg_scat_fn`). That path needs ≥ 200 000 rows as well as > 50 000 groups (`group.c` ~1949); the new `_dg` / `_dl` / `_dan` tables have 120 002 rows and `Tcd-null` in extras has 60 000, so all of them stop at the serial `CD_INSERT` loop, and the comment in `null/distinct.rfl` about "its own serial and partitioned loops" describes coverage that isn't there. I probed it on a release build of this branch with `RAYFORCE_CORES=8`, 240 002 rows alternating `[10 0N]` over 60 001 groups: before `[1]`, after `[2]`; an all-null column `[0]` → `[1]`; F64, DATE and SYM likewise `[2]`. So the fix is right there too, but a regression would pass the suite. The existing `count_distinct.rfl` fixture (`Ncdpg 200000`, `(% (til Ncdpg) 51000)`) is the ready-made shape.

Small things left behind by the change:

- `cdpg_ctx_t.has_nulls` (group.c) and `cdpg_buf_par_ctx_t.has_nulls` (query.c) are now written and never read; `cdpg_hist_fn` still computes `esz` only to `(void)` it.
- The doc comment for the deleted `cdpg_is_null` is orphaned above `cdpg_read` (group.c ~1580).
- In `ray_count_distinct_per_group` the `if (!has_nulls) … else …` split is now two loops computing the same keys, the `else` through `read_col_i64`'s per-row switch, and its comment still mentions a null probe. The typed loops can run unconditionally.
- The new header comment on `cdpg_buf_par_ctx_t` says a null "rides a side flag the same way the literal 0 does", which is only true for SYM; the arm comments beneath say the opposite for I64/I32/I16 and F64. Worth stating the actual rule once, above the if/else chain.
- The F64 arm in query.c now does the `fv != fv` test on no-null columns as well (the "hot path untouched" claim holds for the integer arms, not this one). It is one predicted branch per row, fine, but it is also the eighth verbatim copy of the NaN/-0 fold; a shared `canon_f64_key` in internal.h would make the cross-kernel agreement this PR is about a property of the code rather than of diligence.
- `count_distinct_extras.rfl` still cites `L1204-1227` / `L1223`, which no longer point anywhere.

singaraiona:
One more, non-blocking, found while tracing the other routes: the ungrouped `distinct` builtin (`distinct_vec_eager` → hashset in collection.c ~192) folds NaN into a single null only when `RAY_ATTR_HAS_NULLS` is set; ungated NaNs compare with `==` and each stays distinct. Every grouped kernel now folds NaN unconditionally. So "one rule" holds as long as producers of NaN stamp the attr; arithmetic that yields `0/0` without it would give N at top level and 1 per group. Not introduced here, but it is the last place the rule differs if you want to close it in a follow-up.

PR #470, opened 2026-09-04
feat(ipc): add multicast fan-out

belowzeroff:
## Summary

Adds a core Rayforce multicast module exposed as `.mc.*` over IPC.

Includes:
- topic-based subscribe/unsubscribe/publish
- async `upd(topic, seq, payload)` delivery
- cleanup on IPC close
- aggregate multicast stats
- restricted-mode publish guard
- C regression tests
- real IPC scenario scripts
- usage and implementation docs

## Usage Examples

Subscriber:

```clj
(set upd (fn [topic seq payload]
  (do
    (println "recv topic=% seq=% payload=%" topic seq payload)
    null)))

(set h (.ipc.open "127.0.0.1:5000" 2000))
(.ipc.send h "(.mc.sub \"ticks\" null)")
```

Publisher:

```clj
(set h (.ipc.open "127.0.0.1:5000" 2000))
(.ipc.send h "(.mc.pub \"ticks\" (dict ['sym 'price] (list 'AAPL 185.5)))")
```

Stats:

```clj
(.ipc.send h "(.mc.stats)")
;; {topics:1 subscriptions:1 published:1 delivered:1 dropped:0}
```

## Tests

- `make test TEST_FILTER=mcast`
- real scenarios under `test/scenarios/`


singaraiona:
Design decision on the namespace, after weighing folding this into `.ipc.*`:

**Keep `.mc.*`.** Pub/sub is a different model from point-to-point `.ipc.send`/`.ipc.post` (topics, no handle on the publisher side), and a concern namespace like `.db`/`.csv`/`.ipc` leaves room for replay, retention, or non-IPC transports later without `.ipc` becoming misleading.

**Trim the surface to `.mc.sub`, `.mc.unsub`, `.mc.pub`, `.mc.stats`.** `.mc.drop` duplicates `.ipc.close`: deregistration already calls `ray_mcast_on_close`, so `.ipc.close h` is the eviction path. Please remove `.mc.drop` and document `.ipc.close` instead.

Independent of naming, two of the inline findings need fixing before merge: the fan-out blocking the event loop on a slow subscriber, and list payloads being evaluated on the subscriber instead of quoted.

src/core/mcast.c
```diff
@@ -0,0 +1,326 @@
...
+        return ray_error("domain", ".mc.pub topic is not interned");
+    }
+
+    ray_t* head = ray_sym(ray_sym_intern("upd", 3));
+    ray_t* top  = ray_str(ray_str_ptr(topic_name), ray_str_len(topic_name));
+    ray_t* s    = ray_i64(seq);
+    ray_t* body = payload;
+    if (payload && ray_is_atom(payload) && payload->type == -RAY_SYM) {
+        body = ray_alloc_copy(payload);
+        if (body && !RAY_IS_ERR(body)) body->attrs |= ATTR_QUOTED;
```
singaraiona: **Payload is evaluated as code on every subscriber (remote code execution).** Only a top-level `-RAY_SYM` atom gets `ATTR_QUOTED`; a `RAY_LIST` payload is appended verbatim into `(upd topic seq payload)`, and the subscriber runs `ray_eval` on that list with arguments evaluated (`eval_payload_core` -> `ray_eval`; lambda args at `eval.c` ~3968). `RAY_EVAL_LITERAL_FALLBACK` only rescues lists whose head is *undefined*/non-callable.

Reproduced on this branch: `(.mc.pub "t" (list (first ['println]) "X"))` prints on the subscriber and `upd` receives `null`; `(.mc.pub "t" (list (first ['exit]) 7))` terminates every subscriber process with code 7. Serde only carries `ATTR_QUOTED` on sym atoms (`serde.c:289`), so there is no wire-level way to mark a list literal today — the payload needs to be delivered as data (e.g. a dict message consumed via `.ipc.on.async`, or a receiver-side dispatch that never evaluates argument 3) rather than as an S-expression argument.

src/core/mcast.c
```diff
@@ -0,0 +1,326 @@
...
+    ray_mcast_topic_t* t = ensure_topic(mc, sym);
+    if (!t) return ray_error("oom", NULL);
+
+    int64_t seq = t->next_seq++;
+    mc->published++;
+
+    for (int32_t i = 0; i < t->n_subs;) {
+        ray_t* msg = make_upd_msg(sym, seq, payload);
+        if (!msg || RAY_IS_ERR(msg)) return msg ? msg : ray_error("oom", NULL);
+        ray_err_t rc = ray_ipc_send_async(t->subs[i].handle, msg);
```
singaraiona: **One slow subscriber blocks the whole server.** `ray_ipc_send_async` -> `conn_write_msg` -> `ray_sock_send` waits in `poll(fd, POLLOUT, -1)` on EAGAIN (`sock.c:219-223`), with no timeout, and the `dropped`/`remove_sub` branch only fires on a hard socket error. A subscriber that stops reading fills its kernel send buffer; the next `.mc.pub` then blocks the single poll thread indefinitely — no other client is served, no other subscriber is delivered to, and new connections are refused. Reproduced: a stalled subscriber pinned a 24 MB publish for ~12 s until its socket closed; a third client got connection refused meanwhile. Since `.mc.sub` is `RAY_FN_NONE`, a `-U` restricted client can set this up.

The poll already has a `sel->tx.buf`/`tx.send_fn` slot (currently unused). Fan-out needs a queued, non-blocking tx path with a bounded queue that converts overflow into the drop/remove path, so `dropped` means what the docs say.

src/core/mcast.c
```diff
@@ -0,0 +1,326 @@
...
+    if (ray_is_atom(topic) && topic->type == -RAY_SYM) return topic->i64;
+    if (ray_is_atom(topic) && topic->type == -RAY_STR)
+        return ray_sym_intern(ray_str_ptr(topic), ray_str_len(topic));
+    if (topic->type == RAY_STR && ray_len(topic) == 1) {
+        size_t len = 0;
+        const char* s = ray_str_vec_get(topic, 0, &len);
+        return s ? ray_sym_intern(s, len) : -1;
+    }
+    if (topic->type == RAY_SYM && ray_len(topic) == 1)
+        return ray_read_sym(ray_data(topic), 0, RAY_SYM, topic->attrs);
```
singaraiona: **1-element SYM vector topic bypasses the sym domain and publishes to the wrong topic silently.** `ray_read_sym` returns the raw cell, which is a position in the vector's domain (`sym.h:88-90`), not a runtime sym id; splayed/parted SYM columns carry a FILE domain (`col.c:1484`, `splay.c:493`) and `take` preserves it. Reproduced: `(.mc.pub (take (at t 'sym) 1) 42)` on a splayed table created a second topic under runtime sym id 1 and delivered nothing to the `"ticks"` subscriber, with no error. Use `ray_sym_vec_cell(topic, 0)` and intern its bytes (as the `-RAY_STR` branch does), or `sym_cell_runtime_id`.

src/core/mcast.c
```diff
@@ -0,0 +1,326 @@
...
+ray_t* ray_mcast_pub(ray_poll_t* poll, ray_t* topic, ray_t* payload) {
+    if (!poll) return ray_error("domain", ".mc.pub requires an active poll");
+    int64_t sym = topic_sym(topic);
+    if (sym < 0) return ray_error("type", ".mc.pub topic must be a symbol or string");
+    if (ray_serde_size(payload) <= 0)
+        return ray_error("type", ".mc.pub payload is not serializable");
+
+    ray_mcast_t* mc = poll_mcast(poll);
+    if (!mc) return ray_error("oom", NULL);
+    ray_mcast_topic_t* t = ensure_topic(mc, sym);
```
singaraiona: **Topics are never reaped, and publishing creates them.** `ensure_topic` here allocates a permanent `ray_mcast_topic_t` even when nobody is subscribed; `remove_sub`/`ray_mcast_unsub` never drop a topic when `n_subs` reaches 0, `find_topic` is a linear scan, and `ray_mcast_on_close` walks every topic on *every* selector close. A publisher using dynamic topic names (`(.mc.pub (str "order/" id) x)`, which `docs/multicast-usage.md:221` explicitly allows) grows the table without bound and makes every pub/sub/close O(n_topics) for the life of the process. Suggest `find_topic` + early return in pub, and freeing the topic when its last sub is removed.

src/core/mcast.c
```diff
@@ -0,0 +1,326 @@
...
+    if (!t) return ray_error("oom", NULL);
+
+    int64_t seq = t->next_seq++;
```
singaraiona: **Message is rebuilt per subscriber, and a mid-loop failure leaves a seq gap.** `make_upd_msg` (list alloc, `ray_sym_str`, `ray_str` copy, `ray_sym_intern("upd")`, four appends, optional payload copy) runs once per subscriber although its inputs are loop-invariant; `ray_ipc_send_async`/`conn_write_msg` only borrow `msg`, so it can be built once before the loop. That also fixes the second problem: `next_seq++`/`published++` happen before the loop, and the `return` on a `make_upd_msg` failure at subscriber k means subs 0..k-1 got seq N, subs k.. never do, nothing is counted as `dropped`, and the next publish is N+1 — `published != delivered + dropped` and consumers cannot detect the gap. Build (and fail) before bumping `next_seq`.

src/core/epoll.c
```diff
@@ -151,6 +156,7 @@ void ray_poll_deregister(ray_poll_t* poll, int64_t id)
 
     epoll_ctl((int)poll->fd, EPOLL_CTL_DEL, (int)sel->fd, NULL);
     if (sel->close_fn) sel->close_fn(poll, sel);
+    ray_mcast_on_close(poll, id);
```
singaraiona: **Runs after `close_fn`, so the dying handle is still a subscriber inside the `.ipc.on.close` hook** (same at `kqueue.c:163`). `ipc_on_close` fires `hook_call_lifecycle(IPC_HOOK_CLOSE)` before nulling `sel->data`, so a hook that does `(.mc.pub ...)` writes to the closing fd and records a spurious `delivered` or `dropped`, and `(.mc.stats)` over-counts subscriptions by one.

Also on placement: only ipc.c-registered socket selectors can ever hold a handle, and `ipc_on_close` already runs on every one of those close paths with the same `poll` and `sel->id`. A single `ray_mcast_on_close(poll, sel->id)` at the top of `ipc_on_close` (before the hook) fixes the ordering and removes the `poll -> mcast` coupling duplicated across the two platform backends. The explicit pre-call in `ray_mcast_drop` (`mcast.c:311`) is then also redundant, since `ray_ipc_close` deregisters synchronously.

test/scenarios/mcast_real_negative.rfl
```diff
@@ -0,0 +1,48 @@
...
+(.ipc.send h "(.mc.pub \"orphan\" 99)")
+(set st1 (.ipc.send h "(.mc.stats)"))
+(_mc_assert (== (at st1 'topics) 1) "orphan topic created")
+(_mc_assert (== (at st1 'subscriptions) 0) "orphan subscriptions")
+(_mc_assert (== (at st1 'published) 1) "orphan published")
+(_mc_assert (== (at st1 'delivered) 0) "orphan delivered")
+(_mc_assert (== (at st1 'dropped) 0) "orphan dropped")
+
+(set bad_filter (try (.ipc.send h "(.mc.sub \"bad-filter\" 1)") (fn [e] "nyi")))
+(_mc_assert (== bad_filter "nyi") "bad filter error")
```
singaraiona, line 34: **These scenarios never run, and the negative asserts are tautological.** `test/main.c:770` walks only `test/rfl`; nothing in the Makefile or CI references `test/scenarios/`, and the header on each file points at `test/rfl/system/mcast_real_*.rfl`, which does not exist. Independently, `(try X (fn [e] "nyi"))` returns the constant `"nyi"` for *any* error (verified: a `name` error from a missing builtin passes), so lines 32-39 cannot distinguish the intended error from the builtin being absent. `CONTRIBUTING.md` puts behavioural tests under `test/rfl/<group>/` where the harness's `EXPR !- SUBSTR` marker asserts a specific error code; there is currently zero `.mc.*` coverage in the auto-discovered tree.

src/ops/system.c
```diff
@@ -1519,3 +1520,30 @@ ray_t* ray_ipc_handle_fn(ray_t** args, int64_t n) {
     (void)args; (void)n;
     return make_i64(ray_ipc_current_handle());
 }
+
+ray_t* ray_mc_sub_fn(ray_t** args, int64_t n) {
+    if (n != 2)
+        return ray_error("domain", ".mc.sub expects 2 arguments");
+    return ray_mcast_sub(ray_ipc_active_poll(), ray_ipc_current_handle(), args[0], args[1]);
```
singaraiona, line 1529: **Handle/poll mismatch on the legacy server path.** In `conn_on_payload` (`ipc.c:1139-1145`) `ipc_ctx_set` sets `ipc_handle` to the conn-array index but leaves `ipc_poll` NULL, so `ray_ipc_active_poll()` falls back to the runtime poll and the subscription is stored there under an index that is not one of its selector ids. On `.mc.pub`, `conn_resolve` looks that index up in the runtime poll: either an unrelated IPC selector receives the `upd`, or the send fails and the sub is silently removed; and legacy `conn_close` never calls `ray_poll_deregister`, so the entry otherwise leaks. Not reachable from `rayforce -p`/`\l` (poll path only), but `ray_ipc_server_init`/`ray_ipc_poll` are still exported for embedders. Refusing `.mc.sub` when `ipc_ctx_poll()` is NULL instead of falling back would close it.

singaraiona:
Re-checked the latest head (329d26ed). The deferred-close change and its test look good and cover the reentrancy case. One regression from the fan-out fix is still present and I'd block on it.

**Any payload larger than the free socket buffer disconnects a healthy subscriber.**

`ray_ipc_try_send_async` writes with `MSG_DONTWAIT` and `try_send_bytes` returns failure on the first `EAGAIN`. `ray_mcast_pub` then removes the subscriber and closes its handle. A subscriber that is reading normally gets kicked simply because the kernel cannot absorb a multi-megabyte frame in one `send()`.

Repro against `./rayforce -p PORT` with one healthy subscriber (`h1`, pumping) and a publisher (`h2`):

```clj
(.ipc.send h1 "(.mc.sub \"big\" null)")
(.ipc.send h2 "(.mc.pub \"big\" (til 500000))")   ; ~4 MB
(.ipc.send h2 "(.mc.stats)")
;; {topics:0 subscriptions:0 published:1 delivered:0 dropped:1}
(.ipc.send h1 "(+ 0 0)")                         ; -> h1 connection is closed
```

The subscriber never receives the message and its connection is gone. Small payloads work. `docs/multicast-usage.md:223` now documents this as intended, but it makes multicast unusable for table-sized payloads.

Suggested fix: on `EAGAIN`, queue the unsent remainder (header + payload) into the connection's `sel->tx.buf`, register interest in write-readiness, and flush from a POLLOUT handler in the poll layer. Drop and close only when the queued bytes for that connection exceed a bound derived from an existing budget. That keeps the non-blocking guarantee while letting a healthy subscriber drain at its own pace.

Still open, lower priority: `test/scenarios/*.rfl` are not run by `make test` and the negative `try` asserts return the same string for any error; a connected restricted client can still create topics without bound via `.mc.sub` (reaped on disconnect, so tolerable); `#include <stdio.h>` landed mid-file in `test/test_mcast.c`.

singaraiona:
Re-verified on 9436ed5e. The drain-before-direct-write fix closes the interleaving hole and every earlier repro now passes. Correctness-wise this looks mergeable.

Verified live on the branch:

- ~4 MB publish to a healthy subscriber followed immediately by sync requests on the same handle: reply arrives, all subsequent multi-MB messages delivered, subscription intact.
- Stalled subscriber that never reads: three ~24 MB publishes queued without blocking; a third client's ping and `.mc.stats` returned immediately.
- `make test TEST_FILTER=mcast`: 10/10 pass under ASan/UBSan, no sanitizer output from the server in any run.

Remaining nits, none blocking:

- `test/scenarios/*.rfl` are still not run by `make test`, and the negative `try` asserts return the same string for any error, so they cannot fail on the wrong error.
- Every IPC connection now registers `EPOLLRDHUP`, which was previously checked but never requested. Data is drained before the hangup path so this looks benign, but it is a behaviour change on all connections, not just subscribers.
- `RAY_IPC_TX_MAX_BYTES` is a new hardcoded 256 MB per-connection cap. `docs/multicast-usage.md` describes it as "the IPC frame budget" but no such existing budget exists; worth either wording it as a new per-connection backlog cap or deriving it from an existing limit.

PR #471, opened 2026-09-04
fix(pivot): keep a temporal type when both branches of an if share it

ser-vasilich:
`(if x ts ts2)` over two TIMESTAMP columns answered `TIMESTAMP` when the pool had workers and `I64` when it had none. `DATE` and `TIME` degraded to `I32` the same way. The values are correct either way — only the type the result claims changes.

```lisp
(set T (table [x ts] (list [true false true] (as 'TIMESTAMP [100 200 300]))))
(print (type (at (select {r: (if x ts ts) from: T}) 'r)))
```

| | before | after |
|---|---|---|
| `RAYFORCE_CORES=0` | `'I64` | `'TIMESTAMP` |
| `RAYFORCE_CORES=2` | `'TIMESTAMP` | `'TIMESTAMP` |

Row count makes no difference — I checked at 3 and at 100 000. It is purely whether a worker exists.

## Why

An `if` selects a value, it does not compute one, so two branches of one temporal type yield that type. `promote_if_type` in `graph.c` puts exactly that case ahead of the general numeric widening, which folds a temporal down to its integer storage — right for arithmetic, since the difference of two timestamps is a span rather than an instant, and wrong for a choice between two of them.

`exec_if` has two arms. The eager one runs when workers exist and takes the node's `out_type` as given. The selected one runs otherwise, and re-derives the type from the values it actually materialised; its ladder had no case for two matching temporals, so it read them as storage.

## What is deliberately left alone

The re-derivation is not redundant, and removing it is the obvious wrong fix — I tried it. Deleting `if_selected_type` makes two more helpers unused, which reads like a duplicate of the graph's rule, but the suite disagrees:

- `(if (> v 0) v (sqrt s))` is statically `F64` because of `sqrt`; when the float branch supplies no rows the values are integers, and the selected path narrows to `I64`. Trusting `out_type` gives `15.0` where `15` is expected.
- `(if cond n1 s2)` over mixed numeric and string types picks a different type by value than by static promotion.

So the two arms answer different questions — one "what type can come out", the other "what type did come out" — and only the temporal case needed aligning.

Branches of *different* temporal types are untouched: there is no shared meaning to keep, so `DATE` against `TIMESTAMP` still widens to `I64`, and the test pins that boundary too.

## Note for whoever unifies this later

The temporal predicate now exists three times: file-local in `graph.c`, again in `query.c` as `dag_type_is_temporal` (which additionally unwraps parted types, so the three are not identical), and now in `pivot.c`. Sharing them is a change of a different shape and is not attempted here.

## Tests

Added to `test/rfl/ops/pivot_coverage.rfl` beside the existing `if` coverage: the three temporal types, the values alongside the types so a future narrowing change cannot pass by dropping data, and the different-types boundary.

On the parent commit, run serially, the first case fails with `expected "TIMESTAMP", got "I64"`.

Suite: 3735/3735 with workers. Run with no workers it is 3729/3735, but those six failures are on the parent too — five call `ray_cd_fused` directly and assert non-NULL, though NULL is its documented "decline, use the other path" answer when there is no pool to dispatch to, and the sixth asserts a routing counter for a path that needs workers. Unrelated to this change; worth fixing separately.


singaraiona:
Reviewed on a branch build. The rule added to `if_promote_type` is right as far as it goes, but the new tests never reach it under the default harness, and the worker-dependent type the PR sets out to close is still open in a common case.

**1. The added assertions do not exercise the fix.**

Both branches in the new `pivot_coverage.rfl` cases are bare column scans, so with `TEST_CORES=2` (Makefile default) `exec_if_selected` declines at the trivial-branch gate (`pivot.c:509-517`) and the eager arm answers from `op->out_type`, which was already correct on dev. Verified: parent commit 2ead9e73 with this PR's `.rfl` dropped in passes `RAYFORCE_CORES=2 ./rayforce.test -f pivot_coverage` (and full `make test`). With `RAYFORCE_CORES=0` the file fails earlier, at the pre-existing line 39 `(if x ts ts2) -- 'TIMESTAMP`, which already exposes the bug. To guard it under workers use a non-trivial same-type branch, e.g. `(if x ts (if x ts ts))` (verified failing on dev with `-c 3`), or run the block under a 0-worker pool.

**2. Result type still depends on worker count when one branch gets zero rows.**

`if_eval_branch` returns NULL for `count <= 0`, `if_value_type(NULL)` is 0, and the `a == 0 -> b` shortcut at `pivot.c:326` runs before the new rule, so the selected arm adopts the surviving branch's raw type instead of the graph's promoted `out_type`. With `x` all true: `(if x ts 5)` -> `'TIMESTAMP` at `RAYFORCE_CORES=0`, `'I64` at 2; `(if x ts 5.5)` -> `'TIMESTAMP` vs `'F64`; `(if x d ts)` -> `'DATE` vs `'I64`. The PR's own boundary assertion at line 685 (`(if x d ts) -- 'I64`) flips to `'DATE` serially if the predicate is all-true.

**3. Simpler rule.** `if (a == b) return a;` fixes the same bug with no temporal list: `if_value_type` already normalises F32->F64 and -STR->SYM, and DATE/TIME/TIMESTAMP are the only (t,t) pairs the ladder changes. `if_type_is_temporal` is otherwise a fourth hand-maintained copy of the set (graph.c:424, query.c:312, lang/internal.h:100 plus inline copies in eval.c/agg.c/temporal.c/hll.c). It is also safer for unsupported (t,t) pairs such as GUID, which the ladder currently folds to RAY_BOOL.

**Deeper fix, recommended.** `if_selected_type` (`pivot.c:559`) should trust `op->out_type` whenever both branches produced values and only narrow (F64 -> integer family) when one side is NULL. That deletes `if_promote_type`/`if_type_is_temporal` entirely and closes (2) and (3). `if_promote_type` already diverges from `graph.c` promote elsewhere (SYM: graph folds to I64 then `ray_if` overrides, pivot returns SYM; F32: graph keeps F32+F32, pivot folds to F64), and this PR is the first noticed instance of the drift. The sqrt narrowing the PR calls load-bearing is exactly the one-side-NULL case: `(if (> v 0) v (sqrt s))` is `'I64` at both core counts because sqrt is non-trivial so the selected arm always runs, so `-- 15` still holds.

**Pre-existing, found while verifying (same file, worth doing here):**

- **Parted temporal columns collapse to BOOL.** `promote_if_type`/`type_is_temporal` in `graph.c:428` do not unwrap `RAY_IS_PARTED`, so over a parted table `(if true ts ts)` has static type `'B8` and `(+ (if true ts ts) 1)` returns `[101 201 45 145 245 89]` (timestamps truncated to uint8), both core counts; printing the B8 column aborts under UBSan (`format.c:424`). Unwrap `RAY_PARTED_BASETYPE` as `query.c:312 dag_type_is_temporal` already does.
- **Temporal nulls through `if` lose the nulls attr** on both arms: `(if x d 0Nd)` renders `-5877611.06.22`, `(if x ts 0Np)` renders `1707.09.22D00:12:43...`. `(as 'DATE (list 1 0Ni 3))` prints `0Nd` correctly, so the formatter needs the attr set. The PR extends this to the serial arm, which previously returned I32/I64 and at least printed the raw sentinel.
- **String literal vs STR column**: `if_value_type` maps a `-RAY_STR` atom to SYM before promotion while graph promote makes STR win, so `(if x "hi" s)` is `'SYM` at runtime against a static `RAY_STR` out_type.
- **Comment accuracy**: the new comment says query.c carries "the same predicate" with "three copies"; `dag_type_is_temporal` rejects `t <= 0` and unwraps parted, so it is not the same, and there are more than three copies. Someone following the invitation to unify would either drop the parted unwrap or miss the others.

singaraiona (commented):
**Verdict: correct, recommend merge. One cosmetic fix requested before merging.**

The analysis in the description holds up against the code. `promote_if_type` already had the same-temporal rule and only needed to see through the parted wrapper; the selected arm's private ladder was the thing that drifted, and replacing it with "trust `out_type`, narrow only when a branch supplied no rows and only where the eager arm could not have run" is the right shape: it removes the worker-count dependence without touching the `(if (> v 0) v (sqrt s))` narrowing that the suite pins. The `eager_possible` predicate is computed from the same four conditions the dispatch uses, so the two arms cannot disagree on eligibility. `if_val_has_null` closes a real gap too: an atom null carries no attribute, so the sentinel rendered as a date.

The tests are well built. Forcing the selected arm with a nested `if` so the case cannot pass by accident under `TEST_CORES=2`, pinning the values next to the types, both empty-branch directions, and the parted case that fell through to `RAY_BOOL` all check out.

Please fix before merge:

- `if_type_eager_ok` in pivot.c: the hunk left a stray blank line and de-indented the `/* No RAY_F32: …` comment to column 0 inside the `switch` (visible in the diff around line 493). Purely cosmetic, but it reads as an accident.

Non-blocking:

- `if_val_has_null` returns false for a SYM/STR `out_type`, so `(if x s 0Ns)` with an atom null keeps whatever the previous behaviour was. Probably fine since the empty symbol is in-band, but worth a one-line note or test if you touch this again.
- Agree with your note that the temporal predicate now lives in three places; a shared helper is the natural follow-up.

singaraiona (changes requested):
**Correction to my earlier comment: this needs changes before merge.** A deeper pass that built the PR head and `origin/dev` side by side found regressions I missed. I reproduced the first three myself on a release build of this branch, at `-c 1` and `-c 2` (identical results at both).

**1. Silent truncation when both branches materialise (blocking).** `if_selected_type` now returns the declared `out_type` whenever both branches supplied rows. But `promote()` in graph.c is narrower than the runtime arithmetic for I32/I16/U8/BOOL operands, and the deleted value-based ladder was the only thing widening the result. Non-trivial branches keep the eager arm out, so this is data corruption, not a label change:

```
T: a = I32 [100000 …], b = I16 [30000 …]
(if c (* a a) (* a a))   dev: I64 [10000000000 …]   PR: I32 [1410065408 …]
(if c (+ b b) (+ b b))   dev: I64 [60000 …]         PR: I16 [-5536 …]
```

A widen-only merge (keep `declared` unless `if_value_type` of a live branch is strictly wider) preserves the TIMESTAMP fix and restores this.

**2. Parted SYM column returns dictionary ids (blocking).** `ray_if`'s SYM override in graph.c still compares the raw wrapped branch type to `RAY_SYM`, so after `promote_if_type` unwraps and returns `promote(SYM,SYM) = I64` the override never fires:

```
parted PT, sy = SYM:  (if (> v 2) sy sy)   dev: SYM [a b c d e f]   PR: I64 [1 2 3 4 5 6]
```

**3. Arithmetic over a parted table becomes B8 (blocking).** The unwrap is on `promote_if_type` only; `promote()` used by `+ - * %` still sees wrapped operands and yields `RAY_BOOL`, and with the new trust-the-declared-type policy that is what comes out:

```
parted PT, v = I64:  (if (> v 2) (+ v v) (+ v v))   dev: I64 [2 4 6 8 10 12]   PR: B8 [true …]
```

With an I32 column this aborts under UBSan when printed. The test comment "with a computed predicate the branch types arrive already unwrapped" is not what happens; dev simply never consumed the bad type. Normalising `out_type` at the scan node (`ray_scan` / `ray_scan_table`) to `RAY_PARTED_BASETYPE` would fix 2 and 3 at the root and subsume the unwrap added here.

Reported by the deep pass, not re-verified by me but consistent with the above:

- `eager_possible` passes the wrapped scan type to `if_type_eager_ok`, which has no case for it, so over a parted table the eager arm is never a candidate and the narrowing still runs: `(if (> v 0) v 1.5)` types as I64 over the parted table and F64 over the same data flattened.
- The one-empty-branch narrowing adopts the live branch's raw type, so `(if x (if x d d) ts)` with all-true `x` reports DATE while the new test pins `(if x d ts)` as I64; the header comment says the narrowing is only F64-to-integer.
- `if_val_has_null` only recognises an atom null whose sentinel width equals `out_type`; `(if c f 0N)` with F64 `f` writes -9.22e18 with no null attr. `RAY_ATOM_IS_NULL` already answers the atom question for every type.
- Pre-existing but in a touched function: `exec_if_eager` indexes `cond_p[i]` past a BOOL atom condition, so `(if false f 1)` gives `[1.5 2.5 1.0]`; the new parted test uses a literal condition and passes only because both branches are the same column.

The cosmetic hunk in `if_type_eager_ok` from my first comment still applies.

singaraiona (approved):
**All findings resolved. Approving.**

Re-verified on a release build of `e3306d16` at `-c 1` and `-c 2`, identical answers at both:

| case | before | now |
|---|---|---|
| `(if c (* a a) (* a a))`, I32 a = 100000 | I32 `[1410065408 …]` | I64 `[10000000000 …]` |
| `(if c (+ b b) (+ b b))`, I16 b = 30000 | I16 `[-5536 …]` | I64 `[60000 …]` |
| parted SYM `(if (> v 2) sy sy)` | I64 `[1 2 3 4 5 6]` | SYM `[a b c d e f]` |
| parted `(if (> v 2) (+ v v) (+ v v))` | B8 `[true …]` | I64 `[2 4 6 8 10 12]` |
| `(if false f 1.0)` | `[1.5 2.5 1.0]` | `[1.0 1.0 1.0]` |
| `(if c f 0N)` over F64 | `-9.22e18`, no null attr | `[1.5 0Nf 3.5]` |
| `(if x (if x d d) ts)`, all-true x | DATE | I64, same as the plain form |
| `(if x ts (if x ts ts))` | I64 serially | TIMESTAMP (the original fix, kept) |

The scan-node unwrap is the right home for the parted fix, the widen-only merge with the two named narrowing exceptions is clearly reasoned in the comment, and the NaN-to-int guard closes a UB path I had not flagged. `pivot_coverage.rfl` passes on the new build and pins each row above, including parted-vs-flattened parity. The cosmetic hunk is gone.

PR #481, opened 2026-09-06
fix(csv): infer Rayfall temporal forms

belowzeroff:

A user has a CSV containing temporal values in the same forms Rayfall prints:

    d,ts
    2024.01.02,2024.01.02D01:02:03.004005006

and reads it without an explicit schema:

    (set T (.csv.read "rf_test_csv_rayfall_temporal.csv"))
    (type (at T 'd))
    (type (at T 'ts))

**Before** - the date column was inferred as a symbol instead of DATE:

    test/rfl/io/csv_rayfall_temporal_repro.rfl:3: expected "DATE", got "SYM"  -- src: (type (at T 'd))

**After** - the same values infer and parse as temporal columns:

    (type (at T 'd)) -- 'DATE
    (type (at T 'ts)) -- 'TIMESTAMP
    (at (at T 'd) 0) -- 2024.01.02
    (at (at T 'ts) 0) -- 2024.01.02D01:02:03.004005006

## Root cause / fix

CSV temporal inference now shares the strict typed-ingest parsers. Those parsers accepted the CSV writer's `YYYY-MM-DD` and `YYYY-MM-DDT...` forms, but rejected Rayfall display forms using dotted dates and the `D` timestamp separator.

This keeps the strict full-consumption parser path, but lets `fast_date` accept `YYYY.MM.DD` alongside `YYYY-MM-DD`, and lets `fast_timestamp` accept Rayfall's dotted-date + `D` separator form.

## Tests

    make test TEST_FILTER=csv_rayfall_temporal
    make test TEST_FILTER=csv_types
    make test TEST_FILTER=csv
    make test

### DIFF
--- a/src/io/csv.c
+++ b/src/io/csv.c
@@ -504,14 +504,15 @@ RAY_INLINE int32_t civil_to_days(int y, int m, int d) {
     return (int32_t)(era * 146097 + doe - 719468 - 10957);
 }
 
-/* Strict YYYY-MM-DD: exactly 10 chars, '-' separators at offsets 4/7, and
- * digits in every field.  Mirrors detect_type() and csv_write_date() so that
- * typed .csv.read [DATE] round-trips the writer's output while rejecting
- * malformed separators or trailing garbage (e.g. "2024/01/02", "2024x01x02",
- * "2024-01-02junk") as null instead of silently coercing them to a bogus
- * date.  fast_timestamp() reuses this on the leading date (passing len 10). */
+/* Strict date parser: exactly 10 chars, either YYYY-MM-DD (CSV writer output)
+ * or YYYY.MM.DD (Rayfall display form), with matching separators at offsets
+ * 4/7 and digits in every field.  Reject malformed separators or trailing
+ * garbage (e.g. "2024/01/02", "2024x01x02", "2024-01-02junk") as null instead
+ * of silently coercing them to a bogus date.  fast_timestamp() reuses this on
+ * the leading date (passing len 10). */
 RAY_INLINE int32_t fast_date(const char* p, size_t len, bool* is_null) {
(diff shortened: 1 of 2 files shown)


singaraiona (commented):
**Verdict: correct, recommend merge.**

Verified on a release build of `e0f5d8d1`. Inference goes through `fast_timestamp` / `fast_date` (detect_type's strict full-consumption grammar), so the two parser changes are the whole story, and the numeric scan ahead of them cannot claim a dotted date (a second dot ends the scan before `p == end`).

What I probed, PR vs dev:

| cell(s) | dev | PR |
|---|---|---|
| `2024.01.02` with a blank row | SYM | DATE `[2024.01.02 0Nd]` |
| `2024.01.02D01:02:03.004005006` | SYM | TIMESTAMP, nanoseconds intact |
| `2024-01-02` and `2024.01.02` in one column | SYM | DATE |
| `2024-01.02` (mixed separators) | SYM | SYM, rejected as designed |
| `1.2.3` | SYM | SYM |
| `2024.01` | F64 | F64, unchanged |
| writer output `2000-01-02` / `…T00:00:00.000000100` | round-trips | round-trips |
| typed ingest `[DATE TIMESTAMP]` on dotted input | null | parsed |

One observation, not blocking: the `T`/`t`/space separators are now also accepted after a dotted date (`2024.01.02T01:02:03` infers TIMESTAMP), since only the `D` form is tied to the dotted date. Harmless, and arguably friendlier, but it is a form neither writer produces, so worth a word in the comment if you want the grammar to stay describable.

singaraiona (approved):
Follow-up resolved: `e77c6977` documents the dotted-date + `T`/`t`/space leniency as deliberate, which was my only note. Comment-only change on top of a fix I verified on a build of `e0f5d8d1` (see the table above). Approving.

PR #482, opened 2026-09-06
fix(serde): round-trip builtins by full registered name

belowzeroff:

A user serializes and deserializes a long-named builtin function:

    (de (ser .db.parted.tables))

Before, the encoder wrote only the truncated builtin display name, so decode could not resolve it:

    test/rfl/system/serde_builtin_name_repro.rfl:1: LHS eval error: error: name  -- src: (de (ser .db.parted.tables))

After, deserialize resolves the original builtin. Calling the decoded value now reaches the builtin's own argument validation instead of failing with `name`:

    ((de (ser .db.parted.tables)) 123)
    error: type: .db.parted.tables expects a string db root path, got i64

## Root cause

Builtin function objects inline only the first 13 bytes of their display name. Serde used that inline name for builtin serialization, so long registered names like `.db.parted.tables` and `.sys.querylog.enable` were serialized truncated and could not be looked up during decode.

## Fix

Add an env reverse lookup from builtin object pointer to symbol id and use that in serde so both the size and write passes serialize the full registered env name.

## Tests

- `make test TEST_FILTER=serde_builtin_name_roundtrip`
- `make test TEST_FILTER=serde`
- `make test` (3747/3748 passed, 1 pre-existing skip)

### DIFF
--- a/src/lang/env.c
+++ b/src/lang/env.c
@@ -114,6 +114,20 @@ static struct {
 int32_t ray_env_scope_depth(void) { return __VM ? __VM->scope_depth : 0; }
 int32_t ray_env_global_count(void) { return g_env.count; }
 
+/* Reverse-map a builtin function object to the symbol it is bound under in the
+ * global environment.  Builtins inline only the first 13 bytes of their name
+ * (see fn_set_name / ray_fn_name), so serialization — which must round-trip the
+ * full name — cannot read it off the object.  Every builtin, dotted reserved
+ * names included, is also bound flat in g_env (see reg_bind's ray_env_bind_flat
+ * call), so a pointer scan recovers the full name.  Returns the sym_id, or -1
+ * if `fn` is not a registered global builtin. */
+int64_t ray_env_builtin_sym(const ray_t* fn) {
+    if (!fn) return -1;
+    for (int32_t i = 0; i < g_env.count; i++)
+        if (g_env.vals[i] == fn) return g_env.keys[i];
+    return -1;
+}
+
 /* Query scopes are synthetic and must not eclipse a caller's lexical
  * parameter/let binding.  Other query frames are deliberately ignored so
  * an inner query can bind a same-named column over an outer query column.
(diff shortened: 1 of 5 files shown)


singaraiona (commented):
**Verdict: correct, recommend merge. One note on the rfl test.**

Verified on a release build of `87fc79a9`. The round-trip works for both long names and through a user alias and inside a list; on dev the same expressions fail with `name: deserialize builtin: '.db.parted.ta' not in global environment`. The decoded value is the builtin itself: calling `((de (ser (fn [x] (.db.parted.tables x)))) 5)` reaches `.db.parted.tables`' own argument check. The reverse lookup returns the first binding, which is the registration entry since builtins are bound before any user `set`, and the decoder's `safe_strlen` path already handles arbitrary lengths, so the longer wire string needs nothing on the read side.

The note: the two `-- .db.parted.tables` lines in `serde.rfl` do not test what their comment says. The formatter prints a builtin through `ray_fn_name`, the 13-byte inline copy, so both sides of the comparison render as `.db.parted.ta` and the line would pass on dev too if decode succeeded. Identity is what proves the name survived, and that is covered by `TEST_ASSERT_EQ_PTR` in the C test and by the `!- type` line; I would either drop the two `--` lines or turn them into a call. Related and pre-existing: `(println .db.parted.tables)` prints `.db.parted.ta`. `serde_builtin_name` is exactly the helper the formatter could use to fix that, if you want to fold it in.

Minor: `ray_env_builtin_sym` is a linear scan of the global env per builtin serialised, called twice (size and write). Fine at current sizes; a small cache would be the fix if serialising large builtin-heavy structures ever shows up in a profile.

singaraiona (approved):
Follow-up resolved. `fb17e1c3` replaces the two formatted-output goldens with calls, which is the behavioural proof; I checked both expected codes against the builtins' own argument checks (`.sys.querylog.enable 1 2` → `domain: takes 0 or 1 arguments`, `.db.parted.tables 123` → `type: expects a string db root path`). Test-only change on top of a verified fix. Approving.

PR #492, opened 2026-09-08
fix(store): reject calendar-impossible date partition directories

belowzeroff:

A partitioned table on disk has a directory whose name is shaped like a date but is not a real one — `2024.02.31` (February has no 31st) — next to a valid one:

    db/2024.02.28/t/
    db/2024.02.31/t/

Loading it with `(.db.parted.get "db/" 't)`:

**Before** — `is_date_dir` accepts any day 01–31, so the impossible name is classified as a date partition and `parse_date_dir` normalises it through the civil-calendar formula into a *different real day*. The partition key silently points at the wrong date:

    partition-key column : date
    schema               : [date v]
    partition-key values : [2024.02.28 2024.02.28 2024.02.28 2024.03.02 2024.03.02]
                                                              ^^^^^^^^^^ 2024.02.31 became 2024.03.02

**After** — the impossible day fails date validation, so the set is not treated as date-partitioned; the directory name is kept literal instead of masquerading as a wrong date:

    partition-key column : part
    schema               : [part v]
    partition-key values : [2024.02.28 2024.02.28 2024.02.28 2024.02.31 2024.02.31]

A real leap day (`2024.02.29`) still classifies as a date, so valid dates are unaffected.

## Root cause

`is_date_dir` gated partition directory names with a blanket `day 01–31` check:

    return month >= 1 && month <= 12 && day >= 1 && day <= 31;

so a shaped-but-impossible name like `2024.02.31` passed as a date. `parse_date_dir` then ran it through the civil-calendar formula, which normalises the overflow into a different real day (2024-03-02), silently mislabelling that partition's key.

## Fix

Validate the day against the leap-aware length of its month, reusing `cal.h`'s `MONTHDAYS` / `date_leap_year` — the shared calendar tables the engine already uses for date decode. An impossible name now fails date classification, so `infer_mc_type` falls through to symbol partitioning and the directory name is kept literal.

## Tests

`part.rfl` adds a mixed valid/impossible date set (`2024.02.28` + `2024.02.31`) and asserts it classifies as `'part` (sym) with the partition keys kept **literal** — not `'date` with `2024.02.31` silently read back as `2024.03.02` — plus a real leap day (`2024.02.29`) that must still classify as `'date` so the tightened check does not over-reject. Full `make test`: 3753/3754 pass (1 pre-existing skip).

### DIFF
--- a/src/store/part.c
+++ b/src/store/part.c
@@ -33,6 +33,7 @@
 #include "store/splay.h"
 #include "table/sym.h"
 #include "table/domain.h"
+#include "lang/cal.h"   /* MONTHDAYS, date_leap_year — calendar-date validation */
 #include <string.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -40,8 +41,11 @@
 #include <dirent.h>
 #include <sys/stat.h>
 
-/* Validate YYYY.MM.DD format: exactly 10 chars, dots at pos 4/7,
- * month 01-12, day 01-31. */
+/* Validate YYYY.MM.DD format: exactly 10 chars, dots at pos 4/7, and a real
+ * calendar date.  A blanket day 01-31 check let calendar-impossible names like
+ * 2024.02.31 through; parse_date_dir then silently normalised them into a
+ * different real day (2024.03.02) — a mislabelled partition key.  Validate the
+ * day against the leap-aware length of its month instead. */
 static bool is_date_dir(const char* name) {
     if (strlen(name) != 10) return false;
     if (name[4] != '.' || name[7] != '.') return false;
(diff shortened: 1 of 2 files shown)


singaraiona (approved):
**Correct, approving.** `MONTHDAYS` is cumulative days at month start for both leap variants, so `MONTHDAYS[leap][m] - MONTHDAYS[leap][m-1]` is exactly the length of month `m`, and `date_leap_year` is the same predicate the date decoder uses, so the check cannot disagree with how a valid date is later read. `is_date_dir` is the only classifier (`infer_mc_type` → `parse_date_dir` at partition build), so one site covers the store. The test pins both directions: the impossible day drops the set to `'part` with the names kept literal, and the real leap day still classifies as `'date`.

One non-blocking observation: nothing stops `.db.splayed.set` from creating a `2024.02.31` directory in the first place, so the mislabel is now impossible on read but the odd directory can still be written; a write-side rejection would be the natural companion if it ever matters.

PR #498, opened 2026-09-09
fix(journal): reject explicit writes during auto-journaled eval

belowzeroff:
## Summary
- reject `.log.write` while evaluating an IPC SYNC message that was already captured by auto-journal
- add test coverage for the explicit-write guard

## Tests
- `make test TEST_FILTER=journal/ops_write_auto_journal_eval DEBUG_CFLAGS='-fPIC -Wall -Wextra -Werror -Wstrict-prototypes -Wno-unused-parameter -std=c17 -g -O0 -march=native -DDEBUG -fno-omit-frame-pointer' DEBUG_LDFLAGS=''`

### DIFF
--- a/src/core/ipc.c
+++ b/src/core/ipc.c
@@ -259,6 +259,18 @@ static void ipc_ctx_set(int64_t handle, ray_poll_t* poll) {
     __VM->ipc_poll   = poll;
 }
 
+static _Thread_local bool g_auto_journal_eval = false;
+
+bool ray_ipc_auto_journal_eval(void) {
+    return g_auto_journal_eval;
+}
+
+#ifdef DEBUG
+void ray_ipc_set_auto_journal_eval_for_test(bool enabled) {
+    g_auto_journal_eval = enabled;
+}
+#endif
+
 ray_poll_t* ray_ipc_active_poll(void) {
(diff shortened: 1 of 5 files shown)


singaraiona (approved):
**Approving, with one docs request.** The guard is right: a SYNC payload is already journaled before it runs, so a `.log.write` inside it would persist a second entry alongside the request it came from, and on replay the re-evaluated request would attempt the same write again — either failing with `noopen` mid-replay or appending once more per recovery. Making that an explicit `domain` error at the call site surfaces it where it is written, rather than at the next recovery. ASYNC payloads are not auto-journaled and keep the escape hatch, which is consistent. The thread-local flag is saved and restored around the eval, so nested or failing evals leave it correct.

Two notes:

- `docs/docs/namespaces/log.md` lists `.log.write`'s errors (`noopen`, `type`, `domain` for zero serialised size, ...); please add the new `domain` case and one sentence saying a SYNC request over a journaled server must not call `.log.write`, since the old Phase D test was the only place that documented the escape hatch as usable from a handler.
- Phase D used to be the only positive coverage of `.log.write` round-tripping through replay; the remaining `ops_write_*` tests cover argument errors. Not blocking, but a local-context write-then-replay case would keep the hatch itself pinned.

PR #499, opened 2026-09-09
test(journal): cover snapshot function binding recovery

belowzeroff:
## Summary
- add regression coverage for recovering user-defined function bindings from journal snapshots

## Tests
- `make test TEST_FILTER=journal/snapshot_function_binding DEBUG_CFLAGS='-fPIC -Wall -Wextra -Werror -Wstrict-prototypes -Wno-unused-parameter -std=c17 -g -O0 -march=native -DDEBUG -fno-omit-frame-pointer' DEBUG_LDFLAGS=''`

### DIFF
--- a/test/test_journal.c
+++ b/test/test_journal.c
@@ -1164,6 +1164,36 @@ static test_result_t test_journal_snapshot_empty_env(void) {
     PASS();
 }
 
+/* 7e. Snapshot restores user-defined functions without binding errors. */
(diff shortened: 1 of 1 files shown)


singaraiona (approved):
**Approving.** Test-only: a user-defined function bound before a snapshot survives `ray_journal_recover` and is callable afterwards, checked by overwriting the binding first so recovery has to restore it. Green in CI.

PR #500, opened 2026-09-09
fix(mcast): accept bare subscribe calls

belowzeroff:
## Summary
- allow `.mc.sub topic` as shorthand for `.mc.sub topic null`
- keep explicit filter rejection unchanged
- document the shorthand form

## Tests
- `make test TEST_FILTER=mcast/sub_requires_ipc_context DEBUG_CFLAGS='-fPIC -Wall -Wextra -Werror -Wstrict-prototypes -Wno-unused-parameter -std=c17 -g -O0 -march=native -DDEBUG -fno-omit-frame-pointer' DEBUG_LDFLAGS=''`

Note: the new shorthand assertion is in the socket-backed multicast test; the full local mcast socket subset is blocked in this environment by `ray_ipc_listen(poll, 0)` failing before multicast behavior runs.

### DIFF
--- a/docs/multicast-usage.md
+++ b/docs/multicast-usage.md
@@ -21,7 +21,8 @@ It does not yet provide durable replay, topic filters, wildcard topics, ACLs, pe
 ## API
 
 ```clj
-(.mc.sub topic null)       ;; subscribe the current IPC handle to topic
+(.mc.sub topic)            ;; subscribe the current IPC handle to topic
+(.mc.sub topic null)       ;; explicit no-filter form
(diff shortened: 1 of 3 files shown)


singaraiona (approved):
**Approving.** `.mc.sub topic` as shorthand for the explicit `null` filter, with the arity error kept for zero arguments and the explicit-filter rejection untouched. The socket-backed test covers the shorthand and the zero-arg error, and CI ran the full mcast subset (the local `ray_ipc_listen` failure in your environment is not reproduced there).

PR #530, opened 2026-09-15
fix(store): order integer partitions by value; reject names past int64

belowzeroff:

An integer-partitioned database — say one directory per year, or per id bucket:

```
db/1/t/  db/2/t/  …  db/9/t/  db/10/t/  db/11/t/  db/12/t/
```

**Before** — partitions are visited in string order, so `10`, `11`, `12` sort between `1` and `2`:

```lisp
(at (select {from: P where: (> n 0)}) 'part)   ; [1 10 11 12 2 3 … 9]
(.db.parted.tables "db/")                      ; reads "9" as the newest partition
```

The I64 partition-key column comes back unordered and rows interleave across keys; `.db.parted.tables` and `.db.parted.fill` take the **last** partition as the most recent, so a table first added in partition `12` is invisible to `.db.parted.tables` and never backfilled by `.db.parted.fill`.

**After** — an all-integer set is ordered by value: keys `[1 2 … 9 10 11 12]`, `.db.parted.tables` reads `12`. Two spellings of one value (`01` and `1`) order by name after the value, so the order stays filesystem-independent; they load as equal keys, which `insert` already rejects as `corrupt`.

Separately, a digit-only name that does not fit `int64` (`db/99999999999999999999/`) hit signed-integer overflow while parsing — UB that in a release build wrapped the key to an unrelated number and, under UBSan, aborted:

```
src/store/part.c:109:27: runtime error: signed integer overflow: 999999999999999999 * 10 cannot be represented in type 'int64_t'
```

It is now rejected: `.db.parted.get`, `.db.parted.tables` and `.db.parted.fill` all fail with

```
corrupt: parted db/: partition directory 99999999999999999999: integer name does not fit int64
```

A digit-only name can only mean an integer key, so naming the directory beats silently retyping the whole set to `sym` keys — where `(== part 2)` would match nothing and `(>= part 2)` would return partition `1` too. Beside a non-integer name (`1.5`) the set is `sym`-keyed as before and the long name is just a name, kept literal.

## Root cause

`collect_part_dirs` sorted every partition scheme with `strcmp`. That is the key order for dates (fixed-width `YYYY.MM.DD`) and for opaque names, and not for integers of unequal width — but every consumer takes the collected order as the key order: `ray_read_parted` emits the `MAPCOMMON` keys in it, `ray_parted_tables` / `ray_parted_fill` read `part_dirs[part_count - 1]` as the newest partition.

`parse_int_dir` accumulated `v = v * 10 + digit` with no bound.

## Fix

- `collect_part_dirs` classifies the set once (`infer_mc_type` → DATE / I64 / SYM), sorts it with `qsort` — by value for an integer set, name as tie-break — and hands the key type to `ray_read_parted`, which no longer classifies again. Date and symbol sets sort as before.
- Integer names parse through `ray_parse_i64` (the overflow-checked parser the tokenizer and CSV reader use). A digit-only name past `int64` in an otherwise integer set is `RAY_ERR_CORRUPT` from `infer_mc_type`; `collect_part_dirs` now returns the error object itself (NULL on success) so that case can name the directory — the `io` / `oom` messages are unchanged.

## Tests

- `part.rfl`: partitions `1`, `2`, `10` load with keys and rows in `[1 2 10]` order, `.db.parted.tables` sees a table that exists only in `10`, and an integer range predicate prunes on the ordered keys. A 20-digit name makes `.db.parted.get` / `.tables` / `.fill` fail `corrupt` (aborted under UBSan before); beside `1.5` the same set is `sym`-keyed with the name literal; `9223372036854775807` is an integer key and `9223372036854775808` is `corrupt`.
- `test_lang.c`: `10` / `2` load as `[2 10]` and the immutable-tail contract holds on them (a later key appends, the last key grows, an earlier key is `domain`); `2` / `02` load as equal keys `[2 2]` in name order and `insert` rejects them as `corrupt` — the validator's strictly-increasing branch; the overflow error message names the directory.
- Docs: `db.md` and `functions.md` no longer tell users to zero-pad integer directory names; `.db.parted.get` documents how names pick the key type and the order.

Full `make test` (ASan + UBSan): 3785/3786 pass, 1 pre-existing macOS skip. `fuzz_parse_deep_nesting` needs `ulimit -s` above macOS's 8 MB default under ASan — unrelated to this change.


src/store/part.c
```diff
@@ -63,16 +63,42 @@ static bool is_date_dir(const char* name) {
...
+    bool neg = false;
+    if (*s == '-') { neg = true; s++; }
     if (!*s) return false;
-    for (; *s; s++)
+    /* Accumulate the magnitude negatively: -INT64_MAX-1 is representable,
+     * +INT64_MAX+1 is not, so the negative side covers both signs. */
+    int64_t v = 0;
+    for (; *s; s++) {
         if (*s < '0' || *s > '9') return false;
+        int d = *s - '0';
```
singaraiona: **Design question, decide before merge.** One all-digit directory past int64 silently re-types the *whole* partition set to SYM keys, and nothing names the offending directory. Verified on the built binary with dirs `1`, `2` plus one overflowing name: `(select {from: P where: (>= part 2)})` returns rows from partition `1` too (SYM keys vs int literal), `(== part 2)` matches nothing, and `(insert P 3 ...)` fails with `partition key must be a non-null sym atom, got i64` with no directory named.

This deliberately mirrors #492 (impossible dates -> SYM) and is strictly better than the old UB wrap, so it's a choice rather than a bug. But an all-digit name has unambiguous intent, and `ray_read_parted` already raises `corrupt` errors naming partition dirs. Consider distinguishing "not integer-shaped" (fall to SYM) from "integer-shaped but out of range" (corrupt error naming the dir). The new `Povf` test in part.rfl locks in the silent behaviour.

Two smaller things on this function:
- The negative path (`neg`, the "accumulate negatively" trick, the `INT64_MIN` special-case) is unreachable: the dir filter in `collect_part_dirs` admits only digits and dots, and all callers are fed from that array. A digits-only loop with `if (v > (INT64_MAX - d) / 10) return false;` is the whole fix, and the "possibly with a leading minus" comments here and on `is_integer_str` should say digits only. (If `-` were ever admitted, `INT64_MIN == NULL_I64` would also need rejecting since the insert validator treats it as a null key.)
- `ray_parse_i64` in `src/core/numparse.h` already does overflow-checked int64 parsing and is unit-tested (test_numparse.c) and fuzzed. A differential check showed identical accept/reject and values for every reachable input; the only difference is `+5`, which the filter makes unreachable. `size_t n = strlen(s); if (n == 0 || ray_parse_i64(s, n, &v) != n) return false;` would drop this third copy of the arithmetic.

src/store/part.c
```diff
@@ -187,17 +212,43 @@ static ray_err_t collect_part_dirs(const char* db_root, char*** out_dirs,
...
+        ray_free_raw(part_dirs);
+        return RAY_ERR_OOM;
+    }
+    bool all_int = true;
+    for (int64_t i = 0; all_int && i < part_count; i++)
+        all_int = parse_int_name(part_dirs[i], &int_keys[i]);
     for (int64_t i = 0; i < part_count - 1; i++) {
         for (int64_t j = i + 1; j < part_count; j++) {
-            if (strcmp(part_dirs[i], part_dirs[j]) > 0) {
+            bool after = all_int ? int_keys[i] > int_keys[j]
```
singaraiona: **Small determinism regression.** The swap is on strict `>` with no name tie-break, so equal-valued names (`01`/`1`, `007`/`7`, `0`/`00`) now come out in readdir order. Both pass the digit/dot filter, both parse to the same key, the set stays I64, and `ray_read_parted` has no duplicate-key check, so `(select {from: P})` row order and which dir `.db.parted.tables` / `.db.parted.fill` treat as the last partition can differ between ext4/xfs/tmpfs or after a copy. Pre-PR strcmp was filesystem-independent for every set, and the comment above still says "deterministic order".

Duplicate keys are already degenerate (insert rejects them as corrupt), so low severity, but either add a tie-break:

```c
bool after = all_int ? (int_keys[i] > int_keys[j] ||
                        (int_keys[i] == int_keys[j] && strcmp(part_dirs[i], part_dirs[j]) > 0))
                     : strcmp(part_dirs[i], part_dirs[j]) > 0;
```

or reject non-canonical spellings at classification.

src/store/part.c
```diff
@@ -187,17 +212,43 @@ static ray_err_t collect_part_dirs(const char* db_root, char*** out_dirs,
...
+     * unordered.  Mirrors infer_mc_type: the set sorts numerically exactly
+     * when it will be classified RAY_MC_I64.
      * O(n^2) but partition count is typically small (< 1000 daily partitions). */
+    int64_t* int_keys = (int64_t*)ray_alloc_raw((size_t)part_count * sizeof(int64_t));
+    if (!int_keys) {
+        for (int64_t i = 0; i < part_count; i++) ray_free_raw(part_dirs[i]);
+        ray_free_raw(part_dirs);
+        return RAY_ERR_OOM;
+    }
+    bool all_int = true;
```
singaraiona: **Non-blocking cleanup.** This re-derives the all-integer classification that `infer_mc_type` (same file) already owns, with a parallel `int_keys` heap array, an OOM cleanup block copied from the one above, and a mirrored swap, kept in sync only by the "Mirrors infer_mc_type" comment. A future tweak to `infer_mc_type` (new sub-type, precedence change) would silently reintroduce exactly the unordered-I64-keys bug this PR fixes, with nothing coupling the two. Each name is also parsed three times (here, `is_integer_str`, `parse_int_dir`), and `int_keys` is allocated even for date/sym sets that never read it.

Simpler: `bool by_value = infer_mc_type(part_dirs, part_count) == RAY_MC_I64;` before the sort (optionally returning the mc type via an out-param so `ray_read_parted` stops classifying again), then `qsort` with a comparator as `dir_entry_cmp` in system.c and `cmp_str_ptr` in env.c already do. That removes ~20 lines and replaces the O(n²) exchange sort. Partition count is user-controlled via `csv.parted` (`rows_per_part` has no upper bound), so n log n is cheap insurance, and the "< 1000 daily partitions" comment no longer covers the integer scheme either way.

test/test_lang.c
```diff
@@ -3937,7 +3938,10 @@ static test_result_t test_eval_insert_parted_key_types_impl(const char* root) {
     TEST_ASSERT_NOT_NULL(setup);
     TEST_ASSERT_FALSE(RAY_IS_ERR(setup));
     ray_release(setup);
-    ASSERT_ER_CODE("(insert pi 11 (list 11))", "corrupt");
+    ASSERT_EQ("(at (select {from: pi where: (> id 0)}) 'part)", "[2 10]");
+    ASSERT_EQ("(count (insert pi 11 (list 11)))", "3");
```
singaraiona, line 3948: **Lost coverage.** Dropping the `corrupt` assertion leaves the validator branch `insert: MAPCOMMON partition keys are not strictly increasing` (query.c, ~13075) with no test at all; grep of test/ finds nothing else hitting it (date keys are fixed-width and validated, sym keys are distinct strcmp-sorted names, the hand-built MAPCOMMON tables in test_opt.c never call insert). It is still reachable post-PR: dirs `2/` and `02/` load as I64 keys `[2 2]`, and `(insert P 3 (list 3))` yields exactly that corrupt error (verified). A future change to `parted_key_cells_cmp` or the `cmp >= 0` check would pass silently. Suggest restoring coverage with a fixture saving `2` and `02` and asserting `corrupt`.

singaraiona (commented):
The core fix looks correct: the overflow guard `v < (INT64_MIN + d) / 10` is exact at all four boundaries under UBSan, and the sort key and MAPCOMMON key type cannot disagree since date names contain dots that `parse_int_name` rejects. No blocking correctness bug.

Inline: one design question (overflowing name silently demotes the whole set to SYM), one small determinism regression (no tie-break for equal-valued names like `01`/`1`), one lost test branch, and some cleanup.

**Docs not in the diff:** `docs/docs/namespaces/db.md:84` ("zero-pad integer directory names if their lexical order would otherwise differ from numeric order") and `docs/docs/language/functions.md:293` still instruct users to zero-pad, which is exactly the behaviour this PR removes and contradicts the new part.rfl test (dirs 1/2/10 load as `[1 2 10]`). Both should say integer names are ordered by value, and that names past int64 fall back to symbol keys. The `test/rfl/system/part.rfl:120` comment "reorders dir entries lexically" is now only true for non-integer sets.

src/store/part.c
belowzeroff: Agreed, and done in 22384c1f. `infer_mc_type` now returns `RAY_ERR_CORRUPT` (with the index) when every name is digit-only and one does not fit int64; `collect_part_dirs` turns that into `corrupt: parted <root>: partition directory 99999999999999999999: integer name does not fit int64`, so `.db.parted.get`, `.db.parted.tables` and `.db.parted.fill` all refuse the set instead of retyping it. A mixed set (a digit-only name beside dotted non-date names) stays SYM with the long name kept literal — there every name is an opaque key, so nothing is corrupt. Both cases are in part.rfl; test_lang.c checks the message names the directory.

Also taken: digits only (`is_digits` + `parse_int_dir`, no minus path or comment), and the parse is `ray_parse_i64`: `n > 0 && ray_parse_i64(s, n, out) == n`.

src/store/part.c
belowzeroff: Added the tie-break in 22384c1f: `int_name_cmp` compares values, then `strcmp`, so `01`/`1` and `007`/`7` come out `01 1 007 7` on every filesystem. Kept them as I64 keys rather than rejecting non-canonical spellings — the docs told users to zero-pad until now, so `01..09, 10, 11` sets exist and must keep loading as integers. test_lang.c pins the order (`02`/`2` → rows `[20 2]`).

src/store/part.c
belowzeroff: Done in 22384c1f. `collect_part_dirs` calls `infer_mc_type` once, sorts with `qsort` (`int_name_cmp` / `name_cmp`) and returns the key type through an out-param; `ray_read_parted` no longer classifies again, and the `int_keys` array, the mirrored swap and the copied OOM block are gone. One signature change beyond the out-param: `collect_part_dirs` returns the error object itself (NULL on success), since the corrupt case has to name the directory — the `io`/`oom` messages are unchanged and the three callers shrink to `if (e) return e;`.

test/test_lang.c
belowzeroff: Restored in 22384c1f: `2` and `02` load as I64 keys `[2 2]` (rows `[20 2]`, i.e. name order within the tie) and `(insert pd 3 (list 3))` is `corrupt`.

belowzeroff:
Pushed 22384c1f addressing the review; PR body updated to match.

- **Design:** a digit-only name past int64 in an otherwise integer set is `corrupt`, naming the directory, from every reader of the set (`.db.parted.get` / `.tables` / `.fill`). A mixed set stays SYM with the name literal.
- **Determinism:** equal-valued spellings (`01`/`1`) order by name after the value.
- **Cleanup:** `collect_part_dirs` classifies once via `infer_mc_type`, sorts with `qsort`, hands the key type to `ray_read_parted`; integer names parse through `ray_parse_i64`; no minus path. `collect_part_dirs` returns the error object (NULL on success) so the corrupt case can carry the directory name; `io`/`oom` texts unchanged.
- **Coverage:** `2`/`02` → `[2 2]` and `insert` is `corrupt` (the validator's strictly-increasing branch); overflow message names the dir; mixed-set fallback; `9223372036854775807` loads, `9223372036854775808` is `corrupt`.
- **Docs:** `db.md` (`.db.parted.get` key inference + the insert paragraph) and `functions.md` no longer say zero-pad; the `part.rfl:120` comment now describes key order rather than lexical order.

Full `make test` (ASan + UBSan): 3785/3786 pass, 1 pre-existing macOS skip.

PR #536, opened 2026-09-15
fix(eval): reset call_lambda's cached `self` sym ID with the runtime

ser-vasilich:
## What & why

`call_lambda` binds `self` for a tree-walked call under a sym ID cached in a function-local `static` on first use and never cleared. `ray_runtime_destroy` tears the sym table down, and the next runtime in the same process can intern `"self"` at a different ID. From then on every tree-walked call — every call to a closure, which is never compiled — binds the lambda under the old ID:

- `self` in the body no longer names the call's lambda;
- whatever ordinary name took the old ID is shadowed by the lambda.

A normal single-runtime process is not affected. Anything that destroys and recreates the runtime in one process is: the test runner, and embedders that restart the runtime.

How it showed up: an RFL test doing `upsert` inside a closure called through `map` failed only in the full `make test` run, never alone:

```
upsert: key must be a symbol or integer, got ?
```

The key argument (a parameter of the enclosing lambda) resolved to the inner closure itself. Bisecting the compiled tests that run before the `.rfl` suite narrowed it to one: `compile/let_reserved_name`, whose compile error falls back to the tree walker and fills the cache in its own runtime.

## Fix

Move the ID to file scope and clear it in `ray_lang_destroy`, next to `ray_compile_reset()`, which already resets the compiler's special-form IDs (`env.c` resets its own `self` cache in `ray_env_destroy`). No other sym-ID caches in `src/` outlive the sym table: the remaining one, `*` in `query.c`, is a builtin registered at the same ID in every runtime.

## Test

`runtime/lambda_self_sym_per_runtime` creates two runtimes in turn and pads the sym table by different amounts, so `"self"` gets a different ID in each (the test asserts that premise). In each, a closure that recurses through `self` and returns a captured value must return 42. One cached ID cannot match both runtimes, so the test fails however the cache was filled — by this test or by an earlier one in the suite.

| | new test alone | `runtime/` group |
|---|---|---|
| dev | FAIL: second runtime → `limit: stack overflow` | 42 / 43 |
| this PR | PASS | 43 / 43 |
| this PR, reset line removed | FAIL | — |

Full `make test` (ASan + UBSan): **3814 / 3814** — including, for this check only, the `.rfl` test that exposed the bug (it failed at the same line on dev and passes here; it is not part of this PR). `make release`: no warnings. `make tsan-test`: 20 / 20, no ThreadSanitizer reports.

## Checklist


singaraiona:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Verified along the way: `ray_runtime_destroy` calls `ray_lang_destroy` unconditionally before `ray_sym_destroy`, so the reset is reached on the only teardown path; the reset is the last statement in `ray_lang_destroy` and nothing later in teardown can re-enter `call_lambda`; the compiled path's own `self` cache in env.c and the compiler's `sf_self` were already reset; and the `*` cache in query.c is interned by `ray_register_builtins` at a fixed position in every runtime, so the completeness claim in the description holds.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #535, opened 2026-09-15
Keyed upsert of a table payload writes cell by cell (boxing + append/memcpy per cell), ~5x slower than a column scatter


vbmithr:
## Summary

A keyed `upsert` of a table payload goes through a per-cell path even when the
table-resident key map is attached and every payload column is a typed vector.
Updating a 935-row batch of existing keys in a 360 000-row table takes
0.41 ms. Writing the same values with one `alter … set` per value column takes
0.087 ms, 4.8× less, and that scatter still pays interpreter overhead.

Follow-up to #531 and #533, measured on fdf928bf. In our real-time depth RDB,
this upsert is now the largest remaining cost: about 35% of the CPU per
tickerplant batch.

## Where the time goes

For a table payload, `upsert_apply` (`src/ops/query.c`) does the following:

1. **Boxes every payload cell.** `collection_elem(sc, r, &alloc)` runs for
   each of the m × ncols cells and allocates an atom per cell, which is freed
   at the end. For 935 rows × 6 columns that is 5 610 allocations per call.
2. **Checks types cell by cell** against the target columns.
3. **Hashes and compares each row through those atoms**
   (`upsert_hash_atoms`, `upsert_map_find` → `upsert_row_eq_atoms`).
4. **Overwrites each non-key cell of a matched row the long way**
   (`upsert_write_cell`), for fixed-width columns:
   - `ray_index_drop(&slots[c])`, then clear `RAY_ATTR_SORTED`;
   - `append_atom_to_col(col, cell)`: append at the tail, which may
     reallocate;
   - `memcpy` the tail element into `row`;
   - `res->len = tail`.

   That is one append, one copy and one index-drop call per written cell.

`perf record --call-graph fp`, update mode, N = 360 000, 2 000 batches:

```
 92.0%  upsert_apply              (27.0% self)
 45.7%    upsert_write_cell       (inlined; ray_vec_append 11.8%, memcpy 5.2%)
 12.4%    ray_free
 11.3%    ray_alloc               (+ heap_coalesce 5.1%)
 11.6%    upsert_map_find         (upsert_row_eq_atoms 6.3%)
  3.6%    collection_elem
  1.9%    ray_index_drop
```

Looking up the key map is ~12%. The rest is boxing and cell-at-a-time writing.

## Measurements

`upsert_scaling.rfl` (below), `-c 1`, batch = 935 rows, 200 batches. `levels`
is referenced only by its global binding, so the ukey map is attached and
reused:

[...]

PR #537, opened 2026-09-16
perf(query): write a keyed upsert of a vector payload column by column (#535)

ser-vasilich:
## What & why

Stacked on #536: its fix is what lets this PR's test pass in the full suite, so this targets `fix/lambda-self-sym-stale` and should be retargeted to `dev` once #536 merges (or merged after it).

Follow-up to #535. A multi-row keyed `upsert` boxed every payload cell into an atom, checked types and hashed keys through those atoms, and wrote each matched cell the long way: append to the column, copy the new tail into place, drop the column's index. A 935-row batch over 6 columns paid 5610 allocations and 5610 tail copies per call; in the reporter's depth RDB the upsert was ~35% of the CPU per tickerplant batch.

When the payload is a table (or a row of vectors) whose columns already carry the target's own types, none of that is needed. The batch resolves through the same table-resident key map, the new rows append with one `ray_vec_append_raw` per column, and the matched cells scatter in payload order — index drop and `SORTED` clear once per column instead of once per cell.

Admission requires, per target column, a payload vector of its exact type: BOOL, U8, I16, I32, I64, F64, DATE, TIME, TIMESTAMP, GUID, or SYM already in the runtime domain. STR and LIST columns, partial payloads, integer-width coercions and slices keep the per-cell arm.

The map's slot carries the key's hash beside the row, in the encoding `sym.c` uses for the intern table (`(hash32 << 32) | (row + 1)`). A map that runs out of room is therefore re-placed from its own slots, not dropped and rebuilt by re-reading and re-hashing every key column, and a probe rejects a colliding slot on 32 bits without touching its row. `ukey_fits` splits accordingly: a map that no longer describes the table is still dropped, one that only lacks room is grown. Both slot markers stay reserved and the map is bounded to 2^32 slots; a table at or past 2^32 rows gets no map.

Semantics are unchanged: a new key enters the map at the row it will take, so a key repeated inside one batch still ends at its last value; appends run before the scatter because a repeat can target a row the same batch appended; nulls travel as their sentinels and raise `HAS_NULLS` on the target column.

## Look-ahead, and why it is gated

On a table past the per-core cache almost all of a probe is two misses — the map slot, then the key cells of the row it names — and taken one row at a time they never overlap. The resolve loop therefore exists twice: one instance hashes the batch first and loads both a few rows ahead, the other does neither. The scatter has the same pair, chosen per column by the column's size.

The gate is the probed bytes (slots + key columns) against `UPSERT_PF_MIN_BYTES` (768 KiB), measured on a 512 KiB-L2 machine: a 640 KiB probe set was 3-8% slower with the look-ahead, an 890 KiB one 19% faster. The instances are separate functions with the flag a compile-time constant, and the per-row hash and probe helpers are force-inlined — a runtime flag inside the loops, or letting GCC drop those helpers out of line, each cost more on cached-sized tables than the prefetching gains.

## Numbers

The issue's own script against this PR's base (dev + #536): batch of 935 rows, 200 batches, `-c 1`, five interleaved runs, medians in ms per batch.

| rows in table | base | this PR | |
|---:|---:|---:|---:|
| update 1 000 | 0.451 | **0.034** | 13.4× |
| update 100 000 | 0.543 | **0.045** | 12.1× |
| update 360 000 | 0.720 | **0.071** | 10.2× |
| update 1 000 000 | 0.773 | **0.085** | 9.1× |
| update 1 200 000 | 0.744 | **0.076** | 9.8× |
| insert 1 000 | 0.551 | **0.066** | 8.3× |
| insert 100 000 | 0.622 | **0.086** | 7.2× |
| insert 360 000 | 0.805 | **0.156** | 5.2× |
| insert 1 000 000 | 1.107 | **0.334** | 3.3× |
| insert 1 200 000 | 0.597 | **0.047** | 12.7× |

The reporter's second script — the same 935-row update of a 360k-row table against a column-wise scatter at known row positions (`alter … set` per value column, no key lookup) — now reads the other way round: 0.070 ms for the upsert against 0.211 ms for three `alter … set`, where on the base it was 0.706 against 0.249.

Inserts at 1M gain least of the insert row: that size is where 200 batches cross the map's load factor. The crossing batch is the worst case a growing table pays — 86 ms at 1M rows, the cost of re-placing 1M entries. At 1.2M, where the same 200 batches cross nothing, the insert is 12.7×.

## Test

`test/rfl/table/upsert_vector.rfl` applies every batch twice — once as a vector payload (this path) and once row by row as atoms (the per-cell path) — and requires the two tables to be identical. Batches repeat keys, mix hits with misses, and carry nulls. Covered: the order-book shape from the issue and a second batch riding the map it left; every admitted value type and key type; keys already duplicated in the target; null and NaN keys; the `(list …)` row form; a shared target (writes a copy) and a value target; hash indexes on a key and on a non-key column; the `sorted` attribute; mixing with single-row `upsert` and `delete`; tables past the look-ahead threshold, compared as serialized bytes; and two map growths — one past the capacity, one after a delete has left tombstones, with a key that delete removed appended again through the grown map.

Thirteen sabotages were run against it — no scatter (each instance), no index drop, `SORTED` kept, both `HAS_NULLS` arms, the appended-row key compare, the map entry for a new key, a repeat of an appended key, a wrong hash in the look-ahead pass, and for the growth: dropping live entries, placing them by a wrong slot, and losing the stored hash when a delete patches the map. Each one fails the test.

On this branch (base: dev + #536): full `make test` (ASan + UBSan) **3814 / 3814**, `make tsan-test` **20 / 20** with no ThreadSanitizer reports, `make release` with no warnings.

## Checklist

- [x] PR targets `fix/lambda-self-sym-stale` (#536), not `master` — retarget to `dev` when #536 lands


singaraiona:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Two candidate findings were investigated and dismissed:

1. The SYM arm of `upsert_vec_move_impl` writes `((int64_t*)d)[o] = id` without consulting the destination's adaptive width, which looks like an out-of-bounds write into a W8/W16/W32 column. It is safe: `upsert_apply` declines unless every column passes `upsert_col_ok`, which rejects a SYM column that is not W64 runtime-domain, so a narrow column is re-expressed before this path runs. Verified against a 16-bit SYM column loaded from CSV under ASan: the column is promoted to W64 by the upsert and the values land correctly.

https://github.com/RayforceDB/rayforce/blob/21dba3f1afa80fbd6dea7ee9598db704a26019dd/src/ops/query.c#L15530-L15532

2. `ukey_grow` re-places entries by walking the old slots in slot order rather than row order, which could in principle flip which of two duplicate keys a probe meets first. Not reproducible: with keys duplicated in the target, a grown map and a freshly built one both resolve to the lowest row, and a differential run of the vector path against the per-cell path over twelve batches spanning several growths showed no divergence.

Minor, non-blocking: the `RAY_IDX_UKEY` comment still describes the old slot layout, now that a slot packs the hash in its high half.

https://github.com/RayforceDB/rayforce/blob/21dba3f1afa80fbd6dea7ee9598db704a26019dd/src/ops/idxop.h#L186-L190

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #420, opened 2026-08-20
Journal: crash between snapshot rename and log roll causes double-apply on recovery


singaraiona:
## Summary

`ray_journal_snapshot()` (`src/store/journal.c`) performs two individually-atomic but not jointly-atomic steps:

1. Write the env snapshot to `<base>.qdb.tmp`, then atomically `rename()` it over `<base>.qdb`.
2. Call `ray_journal_roll()`, which renames `<base>.log` to `<base>.<stamp>.log` and opens a fresh empty `<base>.log`.

If the process crashes (power loss, kill) **after** step 1 lands but **before** step 2 completes — or if the roll's rename fails, since that error path deliberately reopens the old log and keeps appending — recovery (`ray_journal_recover`) will load the fresh `<base>.qdb` **and** replay the old `<base>.log`, whose entries are already baked into that snapshot.

Journaled entries are arbitrary evaluated expressions (the raw sync IPC message is logged and re-evaled on replay), so they are not idempotent: `x+:1`, inserts, etc. get applied twice. The corruption is silent — recovery prints a normal-looking "replayed N entries" message.

## Why the current ordering exists

Snapshot-then-roll is the lesser evil: the reverse order (roll first, snapshot second) would, on a crash in between, leave an *old* `.qdb` plus an *empty* live log with the missing entries stranded in a never-replayed archive — i.e. data **loss** rather than duplicate application. But the window itself remains either way.

## Proposed fix: generation-tagged snapshots with an explicit coverage marker

The key requirement: the snapshot must explicitly record which log generations it covers — inferred file freshness (mtime, name ordering alone) is not enough.

- Give the live log an explicit generation counter (rolled archives already carry a stamp identity).
- Snapshot writes `<base>.<gen>.qdb`, recording (in the name or header) "covers everything through log generation G".
- Recovery loads the newest *loadable* snapshot and replays only logs with generation > G.

Then the two renames no longer need to be atomic as a pair: a crash between them means the snapshot says "covers through G" and the live log is still generation G, so recovery correctly skips it.

Retaining N previous generations (3–5) additionally gives a fallback: a corrupt or bug-affected snapshot can be discarded and recovery falls back to gen−1 plus the retained logs after it. Today a bad `.qdb` is fatal with no fallback (the pre-snapshot state was consumed by the roll).

### Minimal variant

[...]

PR #539, opened 2026-09-16
Cover snapshot function restore across server restart

belowzeroff:
Summary:
- add an end-to-end journal regression for a closure-bearing function saved in a snapshot
- restart a fresh server runtime from the same snapshot base and assert the restored function remains callable
- remove the post-snapshot log before restart so the assertion proves qdb restore, not log replay

Tests:
- make test TEST_FILTER=system/log_journal_advanced DEBUG_CFLAGS='-fPIC -Wall -Wextra -Werror -Wstrict-prototypes -Wno-unused-parameter -std=c17 -g -O0 -march=native -DDEBUG -fno-omit-frame-pointer' DEBUG_LDFLAGS=''


singaraiona:
Reviewed the diff and traced the snapshot path. The test is valid and passes for the right reasons — lambdas really do serialize (`RAY_LAMBDA` arms in `src/store/serde.c:253/499/920`), `ray_journal_snapshot` (`src/store/journal.c:832`) emits only user slots, port 19995 / base `rftest_jsnapfn` don't collide with any other test, and bare `.rfl` lines fail on error (`test/main.c:453`) so the unasserted `.log.snapshot` send isn't swallowed. Four things worth addressing:

**1. The stated invariant isn't isolated.** The Phase E comment says the binding "should be callable after the .qdb restore, before any post-snapshot log tail can recreate it" — but the test can't distinguish those two paths. `.log.snapshot` doesn't roll the log; it writes a coverage marker (`journal.c:889`, #420). If that marker regressed, `(set snapshot_inc ...)` would just replay from the log tail and Phase E would stay green. Suggest asserting the snapshot exists and forcing a qdb-only load:

```
(.sys.exec "test -s /tmp/rftest_jsnapfn.qdb") -- 0
(.sys.exec "rm -f /tmp/rftest_jsnapfn.log")
```

before the restart. That turns the claim in the comment into something the test actually enforces.

**2. Largely duplicative of an existing unit test.** `test_journal_snapshot_function_binding_recovers` (`test/test_journal.c:1176`) already covers this — same `(fn [x] (+ x 1))`, same 41 -> 42 assertion — and it *does* prove provenance by overwriting the binding with `0` before `ray_journal_recover`. The genuinely new dimension here is the process boundary (fresh runtime, no pointers carried over from the prior process), which is worth having, but the PR body doesn't frame it that way and as written this is the weaker of the two tests.

**3. Only the closure-free arm is exercised.** `RAY_SERDE_LAMBDA_HAS_CLOSURE` is the fragile branch; `(fn [x] (+ x 1))` captures nothing. A lambda closing over an outer binding would be the higher-value case for an end-to-end restart test.

**4. Minor:** Phase E uses `-L` but never hard-kills — it SIGTERMs. `-l`, as in Phases B–D, tests the same thing faster; `-L` earns its cost only in Phase A where the `kill -9` is the point.

My take: #1 is cheap and turns a near-tautology into a real assertion; #3 would add the coverage that isn't already in `test_journal.c`. With those, happy to see this merge.

belowzeroff:
Addressed the review points:

- made the restart qdb-only by asserting `/tmp/rftest_jsnapfn.qdb` exists and removing `/tmp/rftest_jsnapfn.log` before restart
- changed the saved function to a closure-bearing lambda via `((fn [offset] (fn [x] (+ x offset))) 1)`
- switched Phase E from `-L` to `-l`; Phase A keeps the `-L` durability coverage
- updated the PR body to frame this as process-boundary + qdb restore coverage

singaraiona:
Re-reviewed at `1736fd63`. All four points are addressed, and #1 and #3 are fixed the right way rather than just papered over:

**#1 (provenance)** — checked that `rm -f`ing the log is actually safe: `ray_journal_recover` guards replay with `file_exists(log_path)` (`src/store/journal.c:472`), so an absent log is a clean skip and the `.qdb` branch alone supplies the binding. The `41 -> 42` assertion after restart is now a real test of snapshot restore; if the qdb write or the lambda serde regressed it fails instead of silently falling back to tail replay.

**#3 (closure arm)** — `((fn [offset] (fn [x] (+ x offset))) 1)` builds the inner lambda while `offset` is in scope, so `LAMBDA_CLOSURE` is non-NULL (`src/lang/eval.c:2018`) and the snapshot now routes through `RAY_SERDE_LAMBDA_HAS_CLOSURE` on write plus the "closure must be a dict" validation on read (`serde.c:937`). That's the branch the previous revision skipped and that `test_journal_snapshot_function_binding_recovers` also skips — it's what makes this genuinely additive rather than a re-framing of existing coverage. Withdrawing #2 on that basis.

**#4** — `-l` with Phase A keeping the `-L` durability coverage is the right split.

One leftover nit, non-blocking: the `rm -f ...log` runs before the SIGTERM, while the server still holds the journal `fp`. It works (the inode is unlinked; shutdown's flush/close writes into the doomed inode), but moving it after the exit-wait loop would be tidier and would drop the test's implicit dependence on shutdown never re-creating the path — `open_log_for_append` does get called on a failed roll (`journal.c:827`).

LGTM once CI is green. Right now ubuntu-release, tsan and rayforce-audit have passed; both debug jobs — the ones that actually run `system/log_journal_advanced` — plus fuzz-smoke and static-analysis are still pending, so worth a look at those before merging.

belowzeroff:
Handled the leftover nit too: the test now waits for the server to exit after SIGTERM, then removes `/tmp/rftest_jsnapfn.log` before the qdb-only restart.

Re-ran:

`make test TEST_FILTER=system/log_journal_advanced DEBUG_CFLAGS='-fPIC -Wall -Wextra -Werror -Wstrict-prototypes -Wno-unused-parameter -std=c17 -g -O0 -march=native -DDEBUG -fno-omit-frame-pointer' DEBUG_LDFLAGS=''`

PR #553, opened 2026-09-17
fix(sort,pivot): stable sort on every path; pivot output in first-seen order

ser-vasilich:
## What & why

Three places where the output order depended on how the work was split, with a benchmark for each.

**Sort stability.** The sort contract (`sort.c`, "Total-order comparison of (key, idx) pairs") promises a stable sort in every size regime, and the streaming asof carry relies on it. Three paths broke it:

- packed radix, reverse-sorted shortcut: `indices[i] = n-1-i` mirrors every plateau of duplicates (n > 4096, key ≤ 3 bytes; n = 4096 is stable, 4097 is not). Ties are now emitted run by run from the back, each run in source order; strictly descending input keeps the O(n) reversal — the sortedness pass now counts equal neighbours to tell the two apart.
- STR keys past 64 rows: the in-place American-flag pass moved records inside a bucket with a swap cycle, and the comparison sorts that finish a bucket returned 0 for equal strings, so the shuffle survived. The pass now scatters each byte through a scratch slice in scan order, so equal strings never leave their source order; `strkey_cmp` and the length sort break exact ties on the source row as well. The scratch is one buffer of `n_live` records: sequential sorts allocate it, parallel sorts reuse the drained top-level scatter source.
- STR desc was `reverse(asc)`, which mirrors ties; the single-run shortcut in the wrong direction did the same. Both now emit runs of equal strings in reverse run order, each run in source order, finding run boundaries on the strings themselves — after the radix passes a record's packed parts hold the last repacked window, not the key.

**pivot order.** Past the parallel threshold the ingest partitions groups by hash and fills each partition in task-claim order; pivot handed out its row and column positions in the order it met the groups, so the result's row order and its *column* order depended on the worker count and differed between runs of one binary (values were right). The two walks stay linear over the partitions; each records the earliest source row per pivot value and per index tuple, and one pass at the end ranks both position spaces by that row, permutes the value list and the index-row entries, and repoints every group — the same order group-by gives.

**Measurements** (medians of 5 interleaved fresh processes, `-c 1`, ms; base = `000eeeba`):

| shape | base | this branch |
|---|---:|---:|
| STR asc, 1M rows, 50-word vocabulary | 94.3 | 79.4 |
| STR asc, 1M unique | 111.6 | 100.1 |
| STR desc, 1M, 50-word vocabulary | 113.1 | 83.9 |
| STR desc, 1M unique (equal length) | 135.2 | 146.5 |
| I64 strictly descending, 1M | 5.19 | 5.21 |
| I64 descending with duplicates, 1M | 5.06 | 6.57 |
| I64 random, 1M | 22.7 | 23.7 |
| pivot 200k rows, 1k / 50k groups, default pool | 4.89 / 28.7 | 4.93 / 30.3 |

The stable scatter is cheaper than the swap cycle it replaces, so STR sorts get faster. Two shapes pay for correctness: descending input with ties (runs have to be found instead of mirrored) and descending unique strings of one length (the run walk compares each neighbour's bytes; comparing packed keys instead does not work after a repack). pivot at 50k groups on the pool was 2.0× slower with a sort-based ordering; the linear ranking brings it back to noise.

Tests: `test/rfl/sort/stable_ties.rfl` (18 assertions, 14 of them fail on `dev`), `test/rfl/group/pivot_first_seen_order.rfl` (all 5 fail on `dev` at `-c 8`). On this branch `make test` (ASan/UBSan) 3878/3878, `make tsan-test` 20/20.

## Checklist


singaraiona:
Reviewed and verified the correctness side — both bugs reproduce on `dev` exactly as described (I ran your two new `.rfl` files against a `dev` build: `stable_ties.rfl` fails at line 10 with `[4096 4095 4094 4093]` where stability wants `[4056 4057 4058 4059]`, and `pivot_first_seen_order.rfl` fails at line 8 with `[0 9 1 5 6 7 2 4 3 8]`). On this branch I get 3878/3878 with no ASan/UBSan reports. The reverse-shortcut gating reads correct to me: the O(n) mirror is now taken only when `total_equal == 0`, so there are no ties to mirror, and the all-equal case is caught earlier by the `total_unsorted == 0` identity branch.

One cell in your benchmark table I can't reproduce, and it goes the other way:

> | STR desc, 1M, 50-word vocabulary | 113.1 | 83.9 |

I measure a consistent ~10% **regression** on that shape:

| shape | dev (000eeeba) | this branch |
|---|---:|---:|
| STR asc, 1M / 50-word | 328.8 | **298.1** (faster, matches you) |
| STR desc, 1M / 50-word | 316–330 | **350–352** (slower) |
| I64 desc-with-dups, 1M | 17.6 | 21.0 (slower, matches you) |

Stable over three outer reps — dev 316.1 / 323.6 / 330.3, branch 350.3 / 351.6 / 351.7.

My absolute numbers are ~3x yours, so our shapes are clearly not the same and this may well be a case yours didn't cover rather than a contradiction. What I ran, release builds both sides, `-c 8`, `taskset -c 0-7`, mean of 5:

```clojure
(set N 1000000)
(set vocab (map (fn [i] (format "word%" i)) (til 50)))
(set s (at vocab (% (til N) 50)))
(set T (table [k v] (list s (til N))))
(set R 5)
(println "STR desc:" (/ (timeit (map (fn [_] (xdesc T 'k)) (til R))) R))
```

So: 50 distinct short keys (`word0`..`word49`, 5–6 bytes, varying length), 20k rows per distinct value, input in `% (til N) 50` order rather than pre-sorted.

The mechanism looks plausible to me rather than surprising — STR desc now finds run boundaries by comparing the strings themselves, and a 50-word vocabulary over 1M rows means very long equal-runs, so the run walk does a lot of comparisons. That's the same cost you already call out for the *unique equal-length* desc row (135.2 → 146.5); my shape suggests it also hits the low-cardinality case, which is the one the table currently shows as the biggest win.

Could you re-check that cell with your own harness? If it does regress on low-cardinality desc, the table is the only thing that needs changing — I don't think it affects the correctness argument, and the stability fix is worth a few percent on one shape regardless.

Two small things from the same pass, neither blocking:

- The new pivot ranking is an insertion sort over distinct pivot values (`pivot.c:1451`), so O(pv_count²) with `pv_count` unbounded. Not a new class of problem — the dedup at `pivot.c:1237` is already a linear scan per row — but it adds a comparable quadratic term on a high-cardinality pivot column.
- The cross-task boundary loop reuses `grain = RAY_DISPATCH_MORSELS * RAY_MORSEL_ELEMS` to re-derive dispatch boundaries. That assumption is pre-existing, not yours, but it's now load-bearing for `total_equal` as well as the inversion counts.

Issue #541, opened 2026-09-16
IPC compression threshold is a compile-time constant; on localhost it costs ~30-50% CPU per frame to save 13% bytes


vbmithr:
## Summary

IPC compression cannot be configured. `RAY_IPC_COMPRESS_THRESHOLD` is a
compile-time `#define 2000` (`src/core/ipc.h:34`), so every frame larger than
2 KB is delta+RLE compressed on send and decompressed on receive. That covers
`conn_frame_msg` (async posts, sync sends, multicast publish) and the response
path.

That default suits a WAN client, but it is pure overhead for the common
tickerplant layout: publisher, TP and subscribers on one host or a fast LAN.
There, for a typical tick table, compression saves little on the wire and
costs a lot of CPU on both ends.

**Request:** a knob for this, for example:

- a runtime threshold setting (`0`/`-1` = never compress), global or per
  listener/connection (a `.ipc.open` option, or a listener option that
  multicast topics inherit); and/or
- skipping compression automatically on loopback and UNIX-domain connections.

## Measurement

Two processes on one host, pinned to separate CPUs. A subscriber
(`-c 1 -p PORT`) defines `upd`; a publisher (`-c 1`) sends 20 000 async
`(list 'upd 'depth batch)` posts, each a 914-row depth-shaped table (12
columns: 2 timestamps, 4 SYM, 4 I64, 2 BOOL), then waits until all are
applied. CPU comes from `/proc/<pid>/stat`; wire bytes from summing the
subscriber's `recvfrom` results under `strace` over 1 000 frames. 5
interleaved repetitions, best of 5; every run applied all 18 280 000 rows.

| build | subscriber µs CPU/frame | publisher µs CPU/frame | bytes on the wire/frame |
|---|---:|---:|---:|
| cb6b0371, compression (threshold 2000) | 96.0 | 129.0 | 57 153 |
| cb6b0371 + IPC decode fixes*, compression | 49.5 | 92.0 | 57 153 |
| cb6b0371 + IPC decode fixes*, **never compress** | **25.5** | **64.5** | 65 902 |

\* Prototype patches: batch-interning SYM vectors on decode, and frame buffers
from the buddy heap instead of `mmap`-backed `ray_sys_alloc` (companion issue
to follow). They make the decode path cheap enough to show compression's
share clearly; the last row only changes `RAY_IPC_COMPRESS_THRESHOLD`.

On this data, compression saves **13% of bytes** at the cost of **~48% of the
[...]

Issue #487, opened 2026-09-08
perf(ipc): share serialized multicast frames across subscribers


vbmithr:
## Problem

`ray_mcast_pub` builds one logical `upd` value, but calls `ray_ipc_try_send_async` once per subscriber. Each call independently serializes and potentially compresses the same message in `conn_frame_msg`.

Multicast publication therefore performs O(subscribers x payload size) serialization/compression work and allocation. This is especially expensive for large table/depth batches, even though every unfiltered subscriber receives identical wire bytes.

## Proposal

Serialize and compress an unfiltered multicast publication once, then share the immutable framed bytes among subscriber transmit queues.

Each connection still needs independent delivery state, such as a write offset and queue linkage, but the underlying frame storage can be reference-counted and released after the last connection finishes or is closed.

The design should leave room for future filters: subscribers receiving an identical filtered result could share a frame, while distinct results would use separate frames.

## Motivation

Application-level multicast should avoid repeating payload transformation work for every recipient. Shared frames reduce CPU cost, allocator pressure, and transient/resident memory for tickerplant fan-out while preserving per-connection backpressure and ordering.

## Acceptance criteria

- One unfiltered multicast publication is serialized and compressed at most once, independent of subscriber count.
- Queued subscribers share immutable frame storage but retain independent write offsets.
- Closing, failing, or overflowing one subscriber does not affect another subscriber's frame delivery.
- Shared storage is released after the last reference is consumed or dropped.
- Tests cover immediate and partial writes to multiple subscribers, including one subscriber disconnecting mid-frame.
- A benchmark or instrumentation demonstrates that serialization/compression count no longer scales with the number of unfiltered subscribers.

PR #557, opened 2026-09-17
feat(mcast): one framing per distinct subscriber compression policy

### LINKED ISSUE #551: Multicast publish ignores link-local compression policy (#549 follow-up): local subscribers still pay compress/decompress
## Summary

#549 made compression local to each link (loopback and UNIX-domain peers never compress), but multicast publish still frames with the compile-time `RAY_IPC_COMPRESS_THRESHOLD`:

`src/core/ipc.c` (000eeeba), in the multicast frame builder:

```c
ray_poll_frame_t* frame = conn_frame_msg(msg, RAY_IPC_MSG_ASYNC, 0,
                                         (size_t)RAY_IPC_COMPRESS_THRESHOLD,
                                         &err);
```

The comment there already calls this out: since #487, one frame is shared by every subscriber, so it cannot carry a per-peer policy. Multicast is the tickerplant fan-out path, which #541 was about, so a TP with only local subscribers still compresses on publish, and every subscriber still decompresses.

## Observed

A tickerplant (`.mc.sub`/publish, relay mode) with 2 subscribers on 127.0.0.1, on rayforce 2.6.2.r124.g000eeeba, fed live depth batches (~370 rows per batch, depth-shaped tables). The profile is `perf record -F 999 --call-graph fp` over 30 s:

- TP: `ray_ipc_compress_at` is 2.22% of samples, which is ~3.5% of the TP's own CPU. On the previous build (bd0fcfe2) with ~900 rows per batch, `ray_ipc_compress` was ~6% of the TP's CPU.
- `.mc.stats`: `subscriptions:2 published:941 delivered:1882 framed:941`, which shows one compressed frame per publication.

The #541 benchmark puts compression plus decompression at ~30–50% of per-frame CPU on localhost, for 13% fewer bytes.

## Proposal

Either of the options the code comment suggests:

1. Split a topic's subscribers into at most two framings, local (never compress) and remote (default threshold), building each framing lazily only when a subscriber of that kind exists; or
2. a listener- or topic-level compression threshold that multicast topics inherit, mirroring the `compress` option on `.ipc.open`.

Option 1 keeps #487's one serialization per publication for the all-local case, with no configuration.

singaraiona:
## What & why

Closes #551, reported by @vbmithr as a follow-up to #549.

#549 made compression a per-link decision, but the multicast fan-out frame is shared by every subscriber of a topic (#487), so publish still framed at the compiled-in `RAY_IPC_COMPRESS_THRESHOLD`. A tickerplant with only loopback subscribers compressed on every publication, and every subscriber decompressed — the exact case #541 was about.

## Approach

Option 1 from the issue: bucket a topic's subscribers by policy, build each framing lazily.

`ray_ipc_frame_async_at(msg, threshold, &frame)` frames at an explicit threshold; `ray_ipc_frame_async` keeps its signature and passes the default. `ray_mcast_pub` groups subscribers by the threshold their link reports — `ray_ipc_handle_threshold`, which #549 already added — and builds one framing per distinct value on first use.

| topic | framings | vs before |
|---|---|---|
| all local (tickerplant) | 1, uncompressed | **cheaper** — #487's single serialization preserved, compression dropped |
| all remote | 1, compressed | unchanged |
| mixed | 2 | one extra serialization, not one per subscriber |

`MC_MAX_FRAMINGS` caps a publication at 4 distinct policies; past that everyone shares the first framing. That is *correct* for any peer rather than merely tolerable, because the receiver honours the per-frame `RAY_IPC_FLAG_COMPRESSED` bit — #549 verified this by running a never-compress build against a compressing one in both directions.

`mc->framed` now counts framings built rather than publications, which is what makes the bucketing observable in `.mc.stats` (the issue cites `published:941 ... framed:941`).

## Tests

**`ipc/mcast_local_not_compressed`** — subscribes a raw socket, publishes a 5000-element i64 run (past the threshold, and delta+RLE shrinks it hard), then reads the delivered frame's header and asserts `RAY_IPC_FLAG_COMPRESSED` is clear.

I verified it pins behaviour rather than implementation by reverting **only** `mcast.c` to `dev` and re-running: it fails with `ph.flags & RAY_IPC_FLAG_COMPRESSED != 0 (got 1, expected 0)`, and passes with the change. It also asserts `.mc.pub` reached exactly one subscriber, so the header assertions cannot pass vacuously — while writing it I had a version where the publish silently did nothing and every later assertion was meaningless.

**`ipc/frame_async_at_threshold`** — covers the framing primitive directly: the same payload frames compressed at the default and raw at `NEVER`, asserted on both the flags and the frame size.

`make test`: **3878/3878**, no ASan/UBSan reports.

## Not included

No CPU/byte benchmark of the fan-out win. I built a two-process harness for it and it didn't produce usable numbers in the time I gave it, so rather than quote something I don't trust, the evidence here is the header assertion: local subscribers demonstrably receive raw frames where they previously received compressed ones. @vbmithr's profile in #551 (`ray_ipc_compress_at` at ~3.5% of TP CPU, and every subscriber decompressing separately) is the measured case this removes.

Note the publisher-side saving is the smaller half: the publisher compresses once per publication, but *every* subscriber decompresses, so aggregate savings scale with the fan-out.

## Checklist


singaraiona:
Audit caught a real regression I introduced — pushed a fix.

**The blocking finding is correct.** My framing-failure guard was unreachable (`ferr != RAY_OK` implies `dead_n >= 1`, so `dead_n == 0` never held), and the live path was worse than the base: an allocation failure dropped every affected subscriber, **closed its connection**, and returned `ray_i64(seq)` — a success value — where `ray_mcast_pub` previously returned an error with all subscriptions intact.

The deeper mistake was treating a framing failure like a failed send. A failed send means the *peer* is bad; a framing failure is local and says nothing about any peer. Fixed by building every framing the topic needs **before** the first send, so the failure is detected with nothing delivered: subscriptions are untouched and the publication is all-or-nothing rather than half fanned out. The framings are still built lazily, one per distinct threshold.

Both other observations were right too:

- `framed`'s field comment still said "one per publication, not per subscriber" — corrected to one per distinct policy per publication.
- **The test rationale was wrong in a way worth flagging.** I wrote that `.mc.pub` returns the number of subscribers reached, so `pr->i64 == 1` guarded against a vacuous pass. It returns the topic's *sequence number*, so that assertion passed because `next_seq` starts at 1 — it would not have caught a zero-fanout publication. The real guard is the read that follows: with no publication the socket never becomes readable and the test fails there. Comment corrected and the misleading equality dropped.

On `MC_MAX_FRAMINGS 4`: agreed it is an arbitrary constant with a silent degradation past it and no test. Past the cap everyone shares `framings[0]`, which is wire-correct since the receiver honours the per-frame flag, so I have left it as a follow-up rather than growing this PR.

`make test` 3878/3878, no ASan/UBSan reports.

Issue #542, opened 2026-09-16
IPC decode: SYM vectors interned per cell, and frame buffers mmap per message (prototypes: -48% subscriber CPU/frame)


vbmithr:
## Summary

Receiving a table over IPC has two avoidable costs on master (cb6b0371),
measured on a tickerplant subscriber:

1. **SYM columns are interned cell by cell.** The `RAY_SYM` decode loop in
   `src/store/serde.c` calls `ray_sym_intern` once per cell: hash, take the
   global spinlock, probe, `sym_cache_segments` for an existing sym, unlock.
   Nothing is deduplicated, so a 914-row batch with a `side` column pays 914
   full interns for two distinct strings.
2. **The frame path maps fresh pages per message.** At cb6b0371,
   `ray_sys_alloc` is always an `mmap` and `ray_sys_free` a `munmap`, whatever
   the size. That is the #531 cost, which #532 removed from the engine but not
   from IPC. Per ~66 KB frame:
   - **send** (`conn_frame_msg`, response path): serialized payload,
     compression buffer, send buffer, `ray_ipc_compress`'s delta buffer, and
     the shared `ray_poll_frame`;
   - **receive:** `c->rx_buf` per message, `deser_frame`'s decompression
     buffer, `ray_ipc_decompress`'s decode buffer, and `ray_poll_buf`s.

Prototypes of both fixes, together, cut **subscriber CPU per frame by 48%**
and **publisher CPU per frame by 29%**. Patches below.

Follow-up to #531, #533 and #535. In our production depth subscriber,
deserialization had become the largest single cost: ~29% of CPU, of which
~41% was interning, ~28% decompression (see #541) and ~12% page faults.

## Fix 1: intern a SYM vector as a batch

- New `ray_sym_intern_batch(hashes, strs, lens, n, out_ids)` in
  `src/table/sym.c`: one `sym_lock`, then `sym_intern_nolock` per string
  (`search_lazy = true`, as `ray_sym_intern`).
- The `RAY_SYM` case of `de_raw_inner` decodes in three passes:
  1. lock-free: delimit and hash every cell with the same
     `(uint32_t)ray_hash_bytes`, keeping the existing truncation errors, and
     dedupe into a local open-addressing table; each cell's id slot
     temporarily holds its distinct index;
  2. one `ray_sym_intern_batch` over the distinct strings;
  3. replace each distinct index with its sym id.
[...]

PR #562, opened 2026-09-17
perf(serde): resolve a SYM column's distinct symbols once per encode

### LINKED ISSUE #561: IPC encode: SYM vectors resolved per cell under the global sym spinlock, twice per frame (~25% of a relay TP's CPU)
## Summary

Serializing a SYM vector resolves every cell through the global sym table
**twice, taking the global spinlock once per cell on each pass**. On a
tickerplant that relays tables with `.mc.pub`, symbol-to-string resolution is
now the largest single cost of the fan-out, and the global spinlock is most of
it.

This is the encode-side counterpart of #542, which batched the decode side
(`ray_sym_intern_batch`). Measured on dev at 27cd3982, which already includes
#541/#542/#557.

## Where the time goes

`src/store/serde.c`, the `RAY_SYM` cases of `ray_serde_size` and `ray_ser_raw`
(which must stay in lockstep):

```c
for (int64_t i = 0; i < obj->len; i++) {
    ray_t* s = ray_sym_vec_cell(obj, i);          /* -> ray_sym_str(id)      */
    size += (s ? (int64_t)ray_str_len(s) : 0) + 1; /* out-of-line call       */
}
```

For the runtime domain, `ray_sym_vec_cell` → `ray_sym_domain_str` → `ray_sym_str`,
and `ray_sym_str` does `sym_lock()` / `sym_unlock()` around a single array read:

```c
sym_lock();                       /* atomic_exchange spin loop */
if (id < 0 || (uint32_t)id >= g_sym.str_count) { ... }
if (!g_sym.strings[id] && (uint32_t)id < g_sym.lazy_count) { ... materialize }
ray_t* s = g_sym.strings[id];
sym_unlock();
```

So one frame costs `2 × rows × sym_columns` lock round-trips, plus as many
calls to `ray_str_len` (and `ray_str_ptr` in the write pass). A 10 000-row
depth batch with 4 SYM columns (`publisher instrument action side`) costs 80 000
lock round-trips, for a handful of distinct strings. Nothing is deduplicated:
`side` has two values.

## Measurement

Production tickerplant: `tp_init.rfl` relaying `depth`/`trades`/`mbp1`/`stats`
batches from one loopback publisher to three loopback subscribers. The TP
inserts, `.mc.pub`s and truncates each batch.
`perf record -F 999 --call-graph fp`, 60 s, 1 043 samples:

| inclusive | |
|---|---|
| 38.7% | `ray_mcast_pub` |
| 16.6% | ↳ `ray_serde_size` |
| 10.8% | ↳ `ray_ser_raw` |
| **18.9%** | `ray_sym_str` (both passes) |
| **9.5%** | ↳ `sym_lock` (inlined), i.e. the spin loop itself |
| 5.4% | `ray_str_len` |
| 1.7% | `ray_str_ptr` |

About 25% of the process's CPU (≈ two thirds of the fan-out) goes to turning
sym ids back into strings. That is uncontended: this process is
single-threaded, so the spinlock costs its atomic exchange and store on every
cell even with no one else to contend with.

## Proposed fix

Mirror #542:

1. **Resolve the column once, under one lock.** Add
   `ray_sym_str_batch(const int64_t* ids, int64_t n, ray_t** out)` in
   `src/table/sym.c`: one `sym_lock`, bounds-check and lazy-materialize per id,
   one `sym_unlock`. Or, since `g_sym.strings` is only replaced on growth, make
   the read lock-free the way `ray_sym_domain_str` already is for FILE domains
   (atomically published `(count, strings)`, retired arrays never freed in
   place). That would fix every other `ray_sym_str` caller too.
2. **Deduplicate by id inside the column.** A small id → (ptr, len) cache,
   e.g. direct-mapped on the id, turns `rows` lookups into `distinct`
   lookups. For typical market-data columns (`side`, `action`, `publisher`)
   that is 2–10 lookups per column instead of thousands, and it also removes
   the per-cell `ray_str_len`/`ray_str_ptr` calls.
3. **Don't resolve twice.** `ray_serde_size` and `ray_ser_raw` walk the same
   cells. A resolve pass shared by both (for example, computing the SYM byte
   size once and caching it next to the frame being built in
   `conn_frame_msg` / `mc_frame_for`) halves what remains. This is optional:
   (1) + (2) already make each pass cheap.

The wire format does not change.

I haven't prototyped this. Going by the profile above, (1) + (2) should remove
most of the 25%, so roughly a 20% CPU cut for a relay-style publisher, and the
same for any process whose response frames carry SYM columns.

## Environment

- rayforce dev 27cd3982 (`rayforce-git 2.6.2.r145`), Linux x86-64, built with
  frame pointers
- Loopback links, so compression is already off (#541); `ray_ipc_compress`
  does not appear in the TP profile

singaraiona:
## What & why

Closes #561, reported by @vbmithr as the encode-side counterpart of #542.

`ray_sym_str` takes the global sym spinlock around a single array read, so walking a SYM column cost **one atomic exchange per cell** — and both `ray_serde_size` and `ray_ser_raw` walk it, so a frame paid `2 × rows × cols` lock round-trips for a handful of distinct strings. On a single-threaded tickerplant that spinning is entirely uncontended.

I sized it before starting, by serializing the same 914-row depth batch with its 4 SYM columns and then with those columns as I64:

| | µs/frame | bytes |
|---|---:|---:|
| 4 SYM columns | **104.8** | 65 875 |
| same shape, no SYM | **6.7** | 75 195 |

~94% of encode was symbol resolution — and the SYM version produces *fewer* bytes. The issue's "25% of TP CPU" is a whole-process figure; inside `ser` it was nearly everything.

## Approach

`ray_sym_str_batch` resolves n ids under one lock, with per-id work identical to `ray_sym_str` (bounds check, lazy materialize, same borrowed atoms), so results cannot differ.

`sym_enc_build` dedupes a column's ids lock-free, resolves only the distinct ones through that batch, and **caches each one's `(ptr, len)`**.

That second part came from profiling rather than from the plan. My first version cached only the atom and got −27%; the profile then showed `ray_str_len` (23.2%) and `ray_str_ptr` (7.4%) — out-of-line calls made per cell — were 30% of encode on their own. Caching `(ptr, len)` makes the per-cell work two array loads and took it to −43%.

Non-runtime domains already resolve lock-free, and the `RAY_SYM_AUDIT` hook is a per-cell contract, so both keep the straightforward walk while still populating the same cache, so the hot loops stay uniform.

Scratch is one `ray_alloc_raw` block (buddy heap, not `ray_sys_alloc`), 8-byte array laid out first so every array stays aligned. **No new tunables, no env knobs.**

## Measurements

Release builds, this machine, 914-row depth-shaped batch with 4 SYM columns:

| | before | after |
|---|---:|---:|
| `ser` of the batch | 88.5 µs | **50.2 µs** (−43%) |
| same shape, SYM as I64 (control) | 6.7 µs | 6.5 µs |
| **IPC publisher**, 20k frames | 148.5 µs/frame | **76.0 µs/frame** (−49%) |
| IPC subscriber | 75.5 µs/frame | 70.0 µs/frame |
| bytes on the wire | 65 875 | **65 875** |

Every benchmark run applied all 18 280 000 rows.

## Correctness

The wire format does not change, and the existing SYM round-trip tests cover that dedupe is only for *resolution*: a column of 12 cells with 2 distinct values still encodes 12 strings. Byte counts are identical before and after.

New tests cover `ray_sym_str_batch` against `ray_sym_str` for duplicates, the `""` SYM null, a dotted name, negative and out-of-range ids (NULL, matching `ray_sym_str` rather than aborting the batch), and the empty batch.

`make test`: **3887/3887**, no ASan/UBSan reports.

## Not included

Resolving once for **both** passes — the issue's optional step 3. After this change `sym_enc_build` is 41.7% of encode and the payload `memmove` 19.4%, so sharing one resolve between `ray_serde_size` and `ray_ser_raw` is the remaining lever. Those two are documented as having to stay in lockstep, and coupling them is exactly where a wrong-bytes bug would come from, so it deserves its own change and its own test rather than being bolted on here.

## Checklist


singaraiona:
Audit was right on all four findings — reworked the implementation rather than patching the sizing, which dissolves them together and turned out faster.

**B1 (blocking) was correct and I should have caught it myself.** Sizing every array by row count when the thing being deduplicated is bounded by `ray_sym_domain_count` was simply wrong, and converting a zero-allocation loop into ~24 bytes/row without mentioning it in the commit message compounded that.

Rather than bound the arrays, the encode path now uses a **direct-mapped id → (bytes, len) cache in thread-local storage**, so it allocates nothing at all — the property the original loop had. Sym ids are stable positions, so the cache persists across calls with no per-column setup; a new `ray_sym_epoch` invalidates it on sym-table teardown/re-init, which is the only event that can.

That removes the other three findings as a side effect: no scratch block to leak on the `ray_sym_str_batch` failure path (**N1**), no `int64_t`-over-`const char*` overlay and no aliasing between `ids` and `out` (**N2** — `ray_sym_str_batch` is gone entirely, since nothing called it any more), and no allocation failure, so `ray_ser_raw` cannot return a short write for a SYM column (**N3**).

**One thing the audit didn't raise, which I found by testing a claim I'd written.** My first version's comment asserted the cache "degrades smoothly" at high cardinality. It didn't — a column where every row is a distinct symbol measured **15% slower** than no cache, because it pays the probe and the store on top of a resolve it still has to do. Fixed by sampling the first 256 cells and letting the rest fall through to the original loop verbatim; routing the cold path through a shared helper instead cost another 7% in `ray_ser_raw`, which the profile showed plainly.

Final numbers, release builds:

| | before | after |
|---|---:|---:|
| 914-row depth batch, 4 SYM cols | 90.6 µs | **35.2 µs** (−61%) |
| 20k rows, 3 distinct | 410.2 µs | **128.1 µs** (−69%) |
| 20k rows, 20k distinct (thrashing) | 462.8 µs | 473.9 µs (+2.4%) |
| IPC publisher, 20k frames | 147.0 µs/frame | **46.0 µs/frame** (−69%) |
| bytes on the wire | 65 875 | 65 875 |

Every benchmark run applied all 18 280 000 rows. The residual +2.4% is the bounded sample prefix on a shape that is the opposite of market data.

`make test` 3884/3884, no ASan/UBSan reports.

PR #563, opened 2026-09-17
fix(query): admit a symbol-vector by: to the count-distinct rewrite

ser-vasilich:
## Bug

`try_count_distinct_v2_rewrite` accepted the `by:` clause of a `count (distinct X)` query only as a bare symbol (`by: A`) or a dict (`by: {A: A B: B}`). A composite key written as a symbol vector — `by: [A B]`, the natural form — returned `NULL` from the by-parser, so the query fell to the general path: `ray_group_indices_fn` plus a per-group `exec_count_distinct` with per-atom hashing, ~50% of the time in `ray_alloc`/`heap_coalesce`.

Same query, 300k rows: dict form 11.5 ms, vector form 84 ms. Identical results.

## Fix

A third by-parser branch reads an unquoted SYM vector (1..15 keys) into `K_syms[]` exactly like the dict keys, so the vector form takes the existing two-pass `ray_group` rewrite (inner group by K..+X, outer by K with COUNT), including the `where:`/`desc`/`take:` shapes.

300k rows: 84 → 11 ms; 1.2M rows: 294 → 32 ms.

## Test

`group/count_distinct_vector_by.rfl`: vector form vs dict form vs a per-group oracle; `where` + `desc` + `take` checked against the full filtered result (the taken rows carry the largest counts); three keys.

`make test` 3885/3885, TSan 20/20.


singaraiona:
Reviewed and verified against a local build. **Approve** — the speedup is real and the results are right.

I measured the claim independently, 300k rows / 97×13 composite key / 1009 distinct targets, mean of 5 at `-c 8`:

| | dev | this branch |
|---|---:|---:|
| `by: [a b]` (vector) | 172.4 ms | **27.4 ms** |
| `by: {a: a b: b}` (dict) | 171.8 ms | 28.4 ms |

Identical counts to the dict form, same 1261 rows. Bigger than the 84 → 11 ms in the description, presumably a different shape.

Three notes, none blocking:

- **The comment rewrite at `query.c:3713` asserts something the code doesn't do.** It now says the dict form carries "aliases" to the output, but nothing reads `by_expr` after the parse — `K_names[j] = ray_sym_str(K_syms[j])` uses source syms. So `by: {x: a y: b}` yields columns `[a b u]`, not `[x y u]`. Pre-existing, but the new wording documents the opposite of the behaviour; worth either fixing the aliasing or not describing it.
- **`K_syms[15]` comment.** "Leave room for X in the composite" doesn't match the bound — with `n > 15` rejected, indices 0..14 fill exactly and X is never stored here (the inner key list is a separate `scratch_alloc` of `n_K+1`). No overflow, just misleading for the next reader.
- **Test rationale.** The header says 300k puts it "past the fused kernel's row floor so that path is the one exercised", but `ray_cd_fused` is gated on `n_K == 1 && !where_expr` and every query in the file uses 2 or 3 keys, so the fused kernel never runs — the size only makes the test slow under ASan. Worth adding a single-element `by: [a]` case, which *is* the newly-admitted shape that reaches it.

One finding I want to flag as **unconfirmed**: a reviewer pass claimed the vector form now drops an aliased projection (`z: a`) that `dev` keeps. I could not reproduce that — on my data `(cols (select {from: T by: [a b] z: a u: (count (distinct u))}))` gives `[a b u]` on **both** dev and this branch, so `z` is dropped either way and it looks pre-existing rather than a regression introduced here. If your shape differs I'd be glad to be corrected.

PR #564, opened 2026-09-17
perf(select): evaluate a single-SYM-column group key once per distinct symbol

ser-vasilich:
## Bug

A `by:` expression such as

```
by: (let p (str-find ref "://") (let s (substr ref (+ p 4) -1) ... (substr r 1 sl) ...))
```

over a SYM column compiles into the DAG and runs over every row. Every `substr`/`str-find` over SYM interns its output per row, so a host-extraction key costs ~8 passes of interning over the whole table (`sym_intern_nolock`, `ray_str_ptr`, `memmem` dominate the profile), while the key is a function of the symbol alone.

## Fix

`derived_key_over_sym_domain` (query.c): when the key expression reads exactly one flat SYM column,

1. collect the distinct ids actually present with a first-seen slot table (`pos[id]`),
2. evaluate the expression through the same DAG compiler over that distinct vector — a one-column table adopting the column's domain — so semantics are the compiled ones, not `ray_eval`'s,
3. spread the `du` results over the rows by slot and feed the key to the group node as a constant (`ray_const_vec`).

**Admission.** The evaluation is valid only for a key that is a pure, row-local function of the symbol, so the whole `by:` form has to pass an allowlist walk: heads from `let`/`if`/`cond`, `and`/`or`/`not`, comparison, arithmetic, `abs`/`neg`/`floor`/`ceil`/`round`/`xbar`/`as`, `substr`/`str-find`/`strlen`/`within`/`nil?`/`upper`/`lower`/`like`; operands limited to let-scoped names, literals and the column itself; a head is admitted only while it still means the builtin (a global lambda of that name, which the DAG compiler inlines ahead of the builtin, or a let-bound name of that name rejects it). Anything else — positional operators (`lag`, `differ`, `fills`, `deltas`, moving windows), an aggregate inside the key, a user function, a free name bound in the environment, a quoted symbol that names a table column (the DAG compiler resolves it to the column) — takes the row-wise key as before. Further gates: ≥ 4096 rows; a near-unique probe over the first 65536 rows through a scratch set sized to the probe, before any domain-sized allocation, bails when more than 3/4 are distinct; distinct ≤ rows/2; domain ≤ 16× rows and ≤ 64M; an id outside the domain bails. The null symbol is id 0 and is evaluated once like any other symbol.

**Key column name.** A single computed key on the DAG path took its column name from its op's ext sym, which is not a name (`dev` names every such key `null`; a const node's slot holds the literal). It is now named by the eval-level rule — the last bare symbol of the form, else `key` — whichever key path ran, yields to an output alias spelling the same (`by: (substr ref 0 2) ref: (min ref)` → `[key ref]`), and the rename waits for a lazy result to materialise.

Medians of 3 runs, full query with `where`/`avg`/`min`, base → patched (ms):

| rows | distinct | -c 1 | pool |
|---|---:|---:|---:|
| 300k | 5% | 304 → 50 (0.17) | 323 → 45 (0.14) |
| 300k | 20% | 361 → 75 (0.21) | 376 → 71 (0.19) |
| 300k | 50% | 431 → 104 (0.24) | 438 → 101 (0.23) |
| 1.2M | 5% | 1230 → 168 (0.14) | 1286 → 143 (0.11) |
| 1.2M | 20% | 1390 → 200 (0.14) | 1464 → 171 (0.12) |
| 1.2M | 50% | 1714 → 236 (0.14) | 1719 → 204 (0.12) |

Of the remaining time the key itself is now ~15 ms per 11k distinct symbols plus ~2 ms of spread; the rest is the filter, the group and the aggregates.

All-distinct column (fallback): 0.96–0.99. Keys outside the path (`xbar` over I64, `substr` over STR, an `if` reading two columns): ±3%. Results are byte-identical to the base.

## Test

`group/derived_key_sym_domain.rfl`: result vs an oracle that materialises the key as a column and groups on it — with `where`, a key that keeps the column's own symbols, an all-distinct column (fallback), nulls in the key column, a two-column expression, `differ`/`fills` and a global-vector predicate (all must take the row-wise key; `differ` collapsed to one group before the allowlist), the key column name at 20k rows and below the 4096-row gate, the alias collision, a quoted column name in the key.

`agg_route_stats().key_domain_evals` counts keys evaluated once per distinct symbol; `test_agg_contract/derived_key_per_symbol_route` checks it fires for `substr` and a `let`/`str-find`/`if` key and stays 0 for `differ`, a two-column key, a table below the row gate and a `substr` shadowed by a user lambda — so the path going dead is a test failure, not a silent slowdown.

`make test` and TSan clean.


singaraiona:
Reviewed against a local build of this branch vs `dev` (1388b334). The idea is good and the gates are carefully built, but **this produces silently wrong groups for order-dependent key expressions** — requesting changes.

## Blocking: order-dependent key expressions

```clojure
(set ref (as 'SYMBOL (at ["g" "h" "i" "j" "k" "l" "m" "n"] (% (til 8000) 8))))
(select {from: T by: (lag ref) c: (count v)})
```

| | groups | counts |
|---|---|---|
| dev | **9** | `[1 1000 1000 1000 1000 1000 1000 1000 999]` |
| this branch | **8** | `[1000 1000 1000 1000 1000 1000 1000 1000]` |

`lag` is defined by row order, so evaluating it over the *distinct* vector discards exactly the thing that defines it: the null-seeded first row and the one-row offset both vanish. No error, just a different answer.

The gate admits any expression that reads exactly one column, but `lag`/`lead`/`sums`/`avgs`/`mins`/`maxs`/`fills`/`deltas` all lower into the query DAG and **preserve length**, so the `len != du` guard never fires. The same class covers an expression embedding an aggregate over the column, where the inner `count` sees `du` rather than `nrows`.

The fix needs to be a positive property of the expression — order-independent and row-local — rather than a count of column references.

## Other findings, which I did **not** independently reproduce

Flagging these as second-hand so you can weigh them accordingly:

- **Quoted symbol naming a from-table column.** `collect_col_refs` skips quoted symbols, but `compile_expr_dag` resolves a quoted symbol naming a column to that column; in the one-column mini table it degrades to a symbol atom. Reported as `by: (if (== 'g ref) "same" "diff")` collapsing to a single group. (My variant of this matched dev, so the shape matters.)
- **Unconditional rename of result column 0, no collision check** — the eval-level path has one. Reported: `by: (substr ref 0 2) ref: (min ref)` yields columns `[ref ref c]`, so `(at R 'ref)` resolves to the key rather than the aggregate.
- **The key column name depends on the gates** — `[ref c]` at 8000 rows vs `[null c]` at 100 rows for the same query, and skipped entirely while the result is still `RAY_LAZY`, since the rename runs before `ray_lazy_materialize`.
- **Slot table sized by the domain count is allocated and memset *before* the near-unique probe**, so a query that ends up falling back can pay a >100MB alloc+memset for nothing. Moving the probe ahead of the allocation would avoid it.

## Clean

Refcounting, the fused-WHERE/selection interaction (the const key is full-table length, which is what `exec_group` expects), the parted-table path (parted column types are rejected by the `RAY_SYM` check), and the `int32_t` slot width (bounded by the 64M domain cap).

Happy to re-check once the admission predicate is tightened — the per-distinct-symbol idea is worth having, and on a low-cardinality SYM key it should be a large win.

singaraiona:
Re-reviewed at `fee49827`. **All five findings are fixed — withdrawing the request for changes.**

My earlier comment was against `a61b0be2`; you pushed the fixes after it, so this is a re-check of the current head rather than a correction.

Verified against `dev` (1388b334) with a local build of this branch:

| check | dev | `fee49827` |
|---|---|---|
| `by: (lag ref)` — the blocking one | 9 groups `[1 1000…999]` | **9 groups, identical** ✓ |
| rename collision `by: (substr ref 0 2) ref: (min ref)` | `[null ref c]` | **`[key ref c]`** — no duplicate ✓ |
| key name at 8000 vs 100 rows | `'` / `'` | **`'ref` / `'ref`** — stable across the gates ✓ |
| bare column key `by: ref` | `[ref c]` | `[ref c]` ✓ |
| quoted symbol `by: (if (== 'gg ref) …)` | `[same diff] [1000 7000]` | identical ✓ |

The elementwise allowlist is the right shape for the admission test — a positive property of the expression rather than a count of column references, which is what let `lag` through before.

**The win survived the extra gating, and it is large.** 300k rows, 500 distinct URL symbols, host extracted with `substr`, mean of 5 at `-c 8`:

| | dev | this branch |
|---|---:|---:|
| `by: (substr ref 8 12)` | 243.3 ms | **4.77 ms** |

51×, with identical output (500 rows, 300 000 total). `make test` on the branch: 3886/3886, no ASan/UBSan reports.

One deliberate behaviour change worth calling out in the description if it isn't already: a computed key column is now named after the expression's last bare symbol (`ref`) or `key`, where `dev` produced `null`/`'`. That is an improvement and matches the eval-level path, but it is a visible change to result column names for anyone selecting a computed key by name.

Approving.

PR #568, opened 2026-09-17
fix(table): make the public table accessors total over ray_t

### LINKED ISSUE #567: ray_table_nrows/ncols segfault on a non-table (public API, no documented precondition)
## Summary

`ray_table_nrows()` and `ray_table_ncols()` segfault when handed a value
that is not a table. They are part of the public embedding API in
`rayforce.h`, which takes an opaque `ray_t*` and documents no
precondition, so from an embedder's side these read as total functions.

Related but separate from #565 (which is a `__VM` NULL guard on a
different path).

## Reproduction

Against `dev` at `c27fdc34`. The value is an ordinary scalar reply to
`(+ 1 2)` over IPC — `type=-5`:

```c
ray_runtime_create(0, NULL);
ray_poll_t* p = ray_poll_create(); ray_runtime_set_poll(p);
int64_t h = ray_ipc_connect("127.0.0.1", port, NULL, NULL, 2000);
ray_t* r = ray_ipc_send(h, ray_str("(+ 1 2)", 7));
printf("type=%d\n", r->type);     /* -5 */
ray_table_nrows(r);               /* SIGSEGV */
```

Probing each accessor on that same reply, with a SIGSEGV handler:

```
scalar IPC reply, type=-5:
  table_nrows        -> SIGSEGV
  table_ncols        -> SIGSEGV
  dict_len           -> 0
```

`dict_len` returning 0 is not a guard — `ray_dict_len` is written the
same way (`ray_dict_keys(d)` then `keys->len`); this input just happens
to land on a falsy pointer. Which is the awkward part: **the same call is
benign on one value and fatal on another.** A locally built
`ray_i64(42)` passed to `ray_table_nrows` returns 0 here; only the
IPC-deserialized atom crashes. That makes the misuse very easy to ship —
it is how we found it, in a smoke test that had been passing a scalar
reply to `ray_table_nrows` for a while under a `try ... with` that of
course cannot catch a segfault.

## Why this isn't a complaint about the design

I realise the accessors are written against block layout, not the type
tag, and that this is deliberate: your own call sites check first
(`if (x->type == RAY_TABLE)` in `src/ops/agg.c:546`, the ternary in
`src/core/qlog.c:38`), because the evaluator has already established the
type during builtin dispatch. Keeping `ray_table_nrows` branch-free
matters on paths like `agg_engine.c`'s parallel-size test. And
`ray_table_validate_rectangular()` is there for callers who want an
explicit check.

That works well for the evaluator. It works less well for an embedder,
which has no evaluator in front of the API — `ray_t*` is one type, so
there is nothing to check against except the tag you already store.

Worth noting `ray_table_ncols` *already* returns 0 defensively:

```c
int64_t ray_table_ncols(ray_t* tbl) {
    if (!tbl || RAY_IS_ERR(tbl)) return 0;
```

It rejects NULL and errors but not a wrong tag, which is the case that
actually crashes.

## Possible resolutions, in your preference order not mine

1. **Make them total.** Extend the existing defensive return: add
   `tbl->type != RAY_TABLE` to the guard `ray_table_ncols` already has,
   and give `ray_table_nrows` the same. One predictable, well-branch-
   predicted comparison against a field already in cache. Same for
   `ray_dict_len`.
2. **Document the precondition.** If the branch is unacceptable on hot
   paths, saying so in `rayforce.h` would be enough — I grepped the
   whole header for `caller must` / `precondition` / `undefined` /
   `assumes` and there is currently nothing, so an embedder has no way
   to learn this short of a crash.
3. **Checked variants.** `ray_table_nrows_checked()` or similar, leaving
   the hot path untouched.

Any of these solves it for us; (2) alone would have been enough. We have
guarded it on our side already, so there is nothing urgent here — filing
it because the next embedder will hit the same thing, and because a
public API segfaulting on an in-range value of its own parameter type
seemed worth reporting rather than working around silently.

Happy to send a PR for whichever option you prefer.

singaraiona:
Fixes #567.

## What

`ray_table_nrows` / `ray_table_ncols` guard NULL and errors but not the type tag, then decode the argument's payload as the two-slot `{schema, cols}` layout and dereference whatever comes out.

The reporter framed it as a quirk of IPC-deserialized atoms, since a local `ray_i64(42)` was benign. The real trigger is just payload bytes that dereference badly — a plain I64 vector does it deterministically:

```c
int64_t poison[] = {0x4141414141414141LL, ...};
ray_t* v = ray_vec_from_raw(RAY_I64, poison, 4);
ray_table_ncols(v);   /* SIGSEGV */
```

```
src/table/table.c:274:33: runtime error: member access within misaligned
address 0x4141414141414141 for type 'union ray_t'
```

## Why option 1, and why all eight

These accessors aren't holding a deliberate line that the rest of the layer crosses — they're the outlier in it:

| function | tag check |
|---|---|
| `ray_dict_keys` / `ray_dict_vals` (`dict.c:112`) | `d->type != RAY_DICT` → NULL |
| `ray_parted_nrows` (`table.c:292`) | not parted → `v->len` |
| `ray_table_validate_rectangular` (`table.c:167`) | typed error |
| the eight public table accessors | **none** |

So `ray_dict_len` returning 0 in the reporter's probe was never luck — that path was already total. Making the table side match is consistency, not a new tax.

The issue named two functions; the identical hole is in all eight that touch `tbl_slots`. `ray_table_schema`, `get_col`, `get_col_idx`, `col_name`, `set_col_idx` and `set_col_name` all crash or corrupt the same way, so fixing two would have left six public functions with the reported defect.

## How

One `tbl_is_table()` predicate next to the slot accessors it protects; every public accessor routes through it. Wrong tag → empty answer (`0` / `NULL` / `-1`), mutators → no-ops, no read through a decoded payload.

Also did the reporter's option 2: `rayforce.h` now documents the guarantee, and points callers who want a *diagnosable* rejection at `ray_table_validate_rectangular` — these accessors answer `0`/`NULL` indistinguishably from a genuinely empty table, which is worth saying out loud.

## Tests

Two new cases in `test/test_table.c`, written before the fix and watched to fail:

- `table/accessors_wrong_type` — poisoned I64 vector through all eight, including asserting the mutators left the payload untouched. Before the fix this aborts the suite at `table.c:274`; after, it passes.
- `table/accessors_reject_dict` — a dict shares the table's two-slot layout, so the accessors would read *real* pointers out of it and answer with a dict's shape. The tag is the only thing separating them.

`make test`: **3889 of 3889 passed (0 skipped, 0 failed)**. Nothing else moved, which is the useful signal across all 324 internal call sites — none depended on the unchecked behaviour.

## Performance

The call sites are per-operation, not per-row — `agg_engine.c:128`'s parallel-size test is the hot one the issue worried about, and it runs once per op. The compare hits a field in the same cache line as the NULL check already there.

Measured anyway, interleaved A/B on the 10M splayed store, min-of-3, two passes (ms):

| query | base | fix |
|---|---|---|
| q13 | 86.19 / 86.31 | 84.94 / 85.58 |
| q16 | 83.40 / 84.07 | 86.08 / 81.91 |
| q17 | 92.78 / 92.63 | 92.99 / 92.46 |
| q20 | 60.08 / 59.46 | 62.08 / 59.73 |
| q22 | 76.99 / 76.79 | 76.43 / 77.44 |
| q28 | 375.42 / 374.25 | 373.79 / 376.56 |

Differences fall inside run-to-run spread in both directions.


singaraiona:
Both blocking findings were correct and are fixed in `142de18b`.

**Finding 1 — `ray_table_add_col` inside a totality guarantee it doesn't honour.** Right, and it was the worse half of the two: the accessors crash, `add_col` *writes* — `ray_cow` then `slots[0] = new_schema; slots[1] = new_cols;` straight into the argument's payload. Documenting that as safe to call on an opaque `ray_t*` was the actual defect in the PR, and it was mine: I converted eight functions and then wrote a contract covering eleven.

Took the guard rather than the carve-out. As the finding notes, the shape was already there — it consumes `tbl` and owns its result, so a wrong tag returns a typed `"type"` error and releases `tbl`, exactly as the bad-column path beside it does:

```c
if (tbl->type != RAY_TABLE) {
    ray_t* err = ray_error("type", "table add_col: expected table, got %s",
                           ray_type_name(tbl->type));
    ray_release(tbl);
    return err;
}
```

The header block was the other half of the problem, so it's rewritten: instead of one blanket claim over a list whose members have different return contracts, it now states the wrong-tag answer per function (`0` / `-1` / `NULL` / no-op / typed error) and says explicitly that `add_col` consumes `tbl` on the error path too.

**Finding 2 — coverage narrower than the claim.** Also right, and the omission was self-selecting: `add_col` was the one function I didn't test and the one that would have failed. `table/add_col_wrong_type` was written before the guard and watched to fail in the right place:

```
src/vec/vec.c:238:17: runtime error: member access within misaligned address
0x4141414141414141 for type 'union ray_t'
```

It asserts the typed error, the error code, and that both poisoned payload words survive untouched. `table/accessors_reject_dict` is broadened from three functions to the full set, including `add_col` — your point that a dict carries *real* pointers is what makes that the sharper test of the two.

`make test`: **3890 of 3890 passed (0 skipped, 0 failed)**.

Also took the stale comment from the non-blocking list — `table.h:30` said `RAY_TABLE (13)`, it is 98.

The rest of the non-blocking section matches what I found independently: the ambiguous call sites all gate on the tag first, and the dict-shares-the-layout case was the real regression risk, which is why that test is in the PR.

Issue #572, opened 2026-09-18
read returns empty for size-0 files that have content (/proc, /sys, FIFOs)


vbmithr:
## Summary

`read` returns an empty string for any file that reports `st_size == 0`
but yields data when actually read — every file under `/proc` and
`/sys`, plus FIFOs and character devices. The buffer appears to be sized
from `stat`, so nothing is read and the caller gets `""` rather than an
error.

## Reproduction

`2.6.2.r147.g1388b334`, Linux.

```rfl
(println "uuid    : [" (read "/proc/sys/kernel/random/uuid") "]")
(println "cmdline : [" (read "/proc/self/cmdline") "]")
(println "hostname: [" (read "/proc/sys/kernel/hostname") "]")
(println "fs.size : " (.fs.size "/proc/sys/kernel/random/uuid"))
```

```console
uuid    : [  ]
cmdline : [  ]
hostname: [  ]
fs.size :  0
```

All three files have content:

```console
$ cat /proc/sys/kernel/hostname
sita
$ stat -c '%s' /proc/sys/kernel/hostname
0
```

`.fs.size` reporting 0 is correct — that is what the kernel reports — but
`read` treating it as "nothing to read" is not.

## Why it matters beyond procfs

The failure is silent. There is no error and no short-read indication;
the caller gets a well-formed empty string and continues. Anything that
then branches on the content takes the wrong branch, and any code that
treats "" as a legitimate value inherits a bug that only appears on
special files.

`/proc` and `/sys` are also the normal way to reach host facts an
embedded process may need and that rayforce does not otherwise expose —
kernel entropy, hostname, PID, cgroup limits.

## Suggestion

Read until EOF rather than sizing from `stat`: keep `stat` as a capacity
hint where it is non-zero, but loop on `read(2)` until it returns 0 and
grow the buffer as needed. That also fixes files that grow between the
`stat` and the read, which is a smaller but real race on ordinary files.

Failing that, returning an error for a file whose size cannot be trusted
would at least be loud.

## Context

Found while giving an embedded process a restart identity.
`(read "/proc/sys/kernel/random/uuid")` looked like the obvious source of
entropy and silently produced an empty identity; the working version has
to shell out through a temp file because `.sys.exec` returns only an exit
code. #571 (the other route, `guid`, being identical in every process) is
the same story from a different direction.

PR #576, opened 2026-09-18
feat(sys): process identity in .sys.info; .sys.exec exit codes and output capture

### LINKED ISSUE #573: Expose process/host identity (pid, hostname, wall clock) and let .sys.exec return stdout
## Summary

An embedded rayforce process cannot learn anything about itself — no PID,
no hostname, no wall-clock time — and cannot capture the output of a
command it runs. Two small additions would close both: process/host
identity, and stdout from `.sys.exec`.

## What is reachable today

Looking for a value that is unique to one process start, I checked every
route I could find on `2.6.2.r147.g1388b334`:

| route | result |
|---|---|
| `(.sys.info)` | `{cores:16 page-size:4096 total-mem:49749975040}` — machine, not process |
| `(.sys.args)` | `{file port cores timeit querylog interactive log user source}` — identical on every restart with the same command line |
| `(.sys.build)` | version and build date — identical for every process of a build |
| `(.time.now)` | documented "Monotonic time in milliseconds"; resets on reboot |
| `(date 0)` | day resolution |
| `(guid 1)` | same first value in every process — #571 |
| `(read "/proc/self/...")` | empty, procfs reports size 0 — #572 |
| `(.sys.exec "...")` | exit code only, no stdout |

Every door is closed, including both accidental ones (#571, #572). Fixing
either of those bugs would unblock the specific thing I needed, but the
general gap would remain: a process still cannot say who or where it is.

## Why an embedder wants this

The concrete case. An owner process hands subscribers an
`(epoch, cursor)` pair, where the cursor counts into its own output log
and restarts at zero. A subscriber holding a cursor from before a restart
must be able to tell that the process it is now talking to is a different
one, or a stale cursor looks perfectly valid and it silently resumes
mid-stream. The epoch has to be unique per process start.

With no identity and no entropy, the working version is:

```rfl
(set epoch_path (format "/tmp/rftp-depth-epoch-%-%.txt"
                        (get (.sys.args) 'port) (.time.now)))
(.sys.exec (format "umask 077; set -C; cat /proc/sys/kernel/random/uuid > '%'" epoch_path))
(set epoch_value (first (split (read epoch_path) "\n")))
(.sys.exec (format "rm -- '%'" epoch_path))
```

Three process spawns, a world-readable directory, an interpolated shell
string and a file that leaks if the `rm` fails — to obtain one identity.

Beyond this case, the same absence shows up in ordinary operational work:
logs that cannot name their own host or PID, a process that cannot
include its identity in a heartbeat, and anything that needs to correlate
a rayforce process with what the host's own tooling reports about it.

## How kdb+ handles this

Worth saying up front that kdb+ shares one of the traits here: its RNG
seed is also fixed at startup, which is why `\S .z.p` is such a common
idiom. So determinism is not the difference. The difference is that two
other doors are open, and either one alone is sufficient.

**`system` returns stdout**, not an exit code:

```q
first system "cat /proc/sys/kernel/random/uuid"
```

That is the entire dance above, as one expression, with no temp file, no
noclobber and no cleanup.

**A `.z` namespace exposes the process to itself** — `.z.i` (PID),
`.z.h` (host), `.z.p` / `.z.P` (wall-clock timestamp, nanosecond),
`.z.a` (IP address), `.z.u` (user), `.z.f` (script), `.z.x` (arguments).
So the idiomatic answer needs no entropy at all:

```q
epoch: .z.h,"-",string .z.i,"-",string .z.p    / host, pid, start time
```

That is a *better* identity than a random UUID, because it is
diagnosable: reading it tells you which machine and which process, so an
epoch in a log or a subscriber error is a lead rather than an opaque
token.

Not suggesting rayforce should grow a `.z` namespace — the naming is
q's history, and `.sys` / `.os` are the natural homes here. The relevant
point is narrower: a database that expects to be embedded gets asked
"which process am I?" by every embedder eventually, and kdb+ answering it
directly is why that question never becomes a shell script.

## Minimal proposal

1. **Process and host identity.** PID, hostname and a wall-clock
   timestamp, wherever they fit best — extra keys in `(.sys.info)`,
   an `.os.pid` / `.os.hostname` alongside the existing `.os.getenv`,
   or a wall-clock companion to `.time.now`. Any one of the three is
   useful on its own; PID plus hostname already gives a usable identity
   when combined with a start time.

2. **`.sys.exec` output capture.** Either change its return to include
   stdout, or add a sibling (`.sys.exec.out`, or a variadic form) that
   returns `(exit_code, stdout)`. This is generally useful well beyond
   identity — shelling out is documented as the escape hatch for
   anything `.fs.*` does not cover, and an escape hatch that discards its
   answer only half works.

(2) additionally makes #572 much less sharp, since the common workaround
for a procfs read is to `cat` it.

Happy to send a PR for either if you have a preference on placement and
naming.

singaraiona:
Closes #573.

Two additions, both as decided on the issue: identity as extra keys in `.sys.info`, and `.sys.exec` gaining a real exit code plus an opt-in capture form.

## `.sys.info` carries process identity

```lisp
(.sys.info)
;; => {cores:8 page-size:4096 total-mem:67191541760
;;     pid:48213 hostname:"tp-prod-01"}
```

The reported use case collapses to one expression, with no temp file, no `umask` dance and no cleanup:

```lisp
(format "%-%-%" (get (.sys.info) 'hostname) (get (.sys.info) 'pid) (timestamp 'global))
```

As the issue argues, that beats a random token because it is diagnosable — reading it back names the host and the process.

## `.sys.exec` returns an exit code

It returned the raw `waitpid`-encoded status:

```lisp
(.sys.exec "exit 3")   ;; was 768, now 3
(.sys.exec "false")    ;; was 256, now 1
```

Signals report `128 + signum`, matching the shell. **Success was `0` before and after**, so anything that only tests for zero is unaffected — that covers every `.sys.exec` call in this repo's own test suite, which is why none of them needed touching. Code comparing against an encoded value does need updating, and the docs carry a warning box saying so.

Worth flagging: `docs/namespaces/sys.md` previously documented `(.sys.exec "false") ;; => 256 (waitpid-encoded)` in as many words, while the summary table one screen above said "return its exit code". So this was deliberate exposure of the raw status, not an oversight — I am changing documented behaviour, not fixing a slip.

## `.sys.exec ... 'out` captures stdout

```lisp
(.sys.exec "echo hi" 'out)              ;; => {code:0 out:"hi\n"}
(.sys.exec "echo partial; exit 4" 'out) ;; => {code:4 out:"partial\n"}
```

Code and output come back together rather than either/or. The capture buffer grows geometrically instead of sizing from `stat()`, so pipes and procfs reads come back whole — which is also the usual workaround for #572, as the issue notes.

**stderr is deliberately not captured.** A shelled-out command's diagnostics belong on the process's own stderr and in the log; merging them into the returned value would corrupt anything parsing it. There is a test asserting this.

## One premise in the issue was wrong

The table lists no wall-clock source, but `(timestamp 'global)` already returns UTC wall clock:

```
2026.09.18D10:05:59.000000000
```

It is built on `time(NULL)`, so it is second-resolution — which is presumably why it did not look like a usable epoch input. With `pid` now available that is sufficient, since pid separates two starts inside the same second. Raising `ray_timestamp_now_ns` to `CLOCK_REALTIME` would give true nanoseconds, but it also feeds the query log, so I have left it as a separate question rather than bundling a timestamp-resolution change into this PR.

## Tests

`test/rfl/system/process_identity.rfl`, written before the implementation and watched to fail. Covers:

- pid/hostname type, positivity, non-empty, and stability within a process
- the machine keys still present
- exit codes `0` / `3` / `42` / `127` (the shell's "command not found", which would otherwise read as `32512`)
- code and output reported together
- empty output is `""`, not null or a missing key
- 100 KB of output — larger than one read buffer — returned whole
- stderr not appearing in the capture
- `'nope` and a non-symbol both rejected

Two existing tests asserted `(count (.sys.info))` is 3; updated to 5.

`make test`: **3891 of 3891 passed (0 skipped, 0 failed)**. `mkdocs build --strict` clean.

Docs updated: `namespaces/sys.md` for both builtins, and the `reference/all-functions.md` rows.


singaraiona:
Blocker fixed in `9d1e7e4a`, and I took three of the non-blocking observations as well.

**The blocker was correct and the reachability argument was exactly right.** `exec_capture`'s only reference lives in the `#else` arm of the `RAY_FUZZING` switch, so `-DRAY_FUZZING` leaves it with zero references and `-Wunused-function` under `-Werror` takes the fuzz build down. Your parenthetical was right too — `exec_exit_code` survives only because `exec_capture` references it, so it alone would not have warned. Both helpers are now inside `#ifndef RAY_FUZZING`.

Reproduced and confirmed locally with `RAY_MARCH=x86-64 make fuzz-smoke`:

```
src/ops/system.c:830:15: error: unused function 'exec_capture' [-Werror,-Wunused-function]
```

and after the fix, all three fuzz targets build and complete.

**`system()` returning -1 — taken.** This was the sharpest of the three: `-1` decoded to `128 + 127 = 255`, indistinguishable from a command that genuinely exited 255, and the capture path was already handling the analogous `pclose` case properly. It is now an `io` error on that path too, and I added `(.sys.exec "exit 255") -- 255` to the tests so the two cases stay provably distinct.

**Allocation-error checks — taken.** The `{code, out}` build now guards both `ray_sym_vec_new` and `ray_list_new`, releasing whatever was already built, matching `ray_sysinfo_fn` beside it.

**Both test observations — taken, and you were right about what they were testing.**

The 100 KB producer leaned on `seq` and on `ARG_MAX` rather than on the buffer growth it was meant to exercise; it is now `head -c 100000 /dev/zero | tr '\0' x`, which is a pipe and therefore also closer to the case that motivated the capture path.

The hostname assertion was the better catch: asserting non-empty tests something `ray_os_hostname` explicitly does not promise, since it degrades to `""` by design. Rather than weaken it to a vacuous check, it now cross-checks against `uname -n`, which tests what actually matters — that when a hostname *is* reported, it is the correct one.

**On the Windows key count:** agreed it is pre-existing and the `.rfl` suite is POSIX-targeted, so I have left it. Worth recording what the arithmetic now is, since the audit note stops just short: `pid` and `hostname` are appended outside the `#if`, so Windows goes from 1 key to 3 while POSIX goes from 3 to 5. The asymmetry is unchanged in kind — the Windows arm still lacks the `sysconf`-backed machine facts — and widening it was not the intent.

`make test`: **3891 of 3891 passed**. `make fuzz-smoke`: clean.

PR #577, opened 2026-09-18
fix(io): read and read-bytes read until EOF, not to the reported size

### LINKED ISSUE #572: read returns empty for size-0 files that have content (/proc, /sys, FIFOs)

singaraiona:
Closes #572.

## Cause

`read_file_bytes` sized the read from `fseek`/`ftell` and then required `fread` to return exactly that count. Anything whose reported size is not its content length broke.

Reproduced on current `dev` — the reporter's own script, verbatim:

```
uuid    : [  ]
cmdline : [  ]
hostname: [  ]
fs.size :  0
```

and after:

```
uuid    : [ 2510e487-aa03-490e-b60f-a6aa16dcc8f1 ]
cmdline : [ ./rayforce ]
hostname: [ Ubuntu-2404-noble-amd64-base ]
fs.size :  0
```

`.fs.size` still reports `0`, which as the issue says is correct — that is what the kernel reports.

## Two more cases with the same cause

The issue names procfs; the diagnosis also covers two things it mentions only in passing.

**FIFOs were not silently empty — they were an outright error.** `fseek` fails on a pipe, so `(read "/some/fifo")` returned `io` before reading a byte. Reading to EOF makes a FIFO behave like any other stream, and there is a test for it.

**The size/read race was a spurious failure, not a truncation.** Because the old code required `rd == sz`, an ordinary file that changed between the measure and the read raised `io` rather than returning short. The issue called this "a smaller but real race"; worth recording that its symptom was a false error.

`read-bytes` shares the helper and was equally affected.

## Fix

Loop on `fread` until EOF, growing as needed. The reported size is kept as a capacity hint where it is plausible, so the ordinary case is still one allocation and one read.

**Measured on a 90 MB file, three runs each:**

| | time | peak RSS |
|---|---|---|
| before | 0.04 s | 102332 / 102336 / 102332 KB |
| after | 0.04 s | 102332 / 102204 / 102336 KB |

Unchanged. The `+1` on the hinted capacity is what lets a single `fread` distinguish "exactly the hinted size" from "more than the hint", so a file that grew is read whole rather than truncated.

## A directory seeks fine and reports LONG_MAX

UBSan caught this during the first test run:

```
src/ops/builtins.c:2069:48: runtime error: signed integer overflow:
9223372036854775807 + 1 cannot be represented in type 'long int'
```

An implausible hint is now treated as no hint. The read itself is what rejects a directory — `fread` sets the error flag (`EISDIR`) and we return `io` — which is also what the test asserts.

## Tests

`test/rfl/io/read_until_eof.rfl`, written before the fix and watched to fail at the `/proc/sys/kernel/hostname` case.

- ordinary, genuinely empty, and 300 KB files (the last exercises the growth loop)
- a genuinely empty file still returns `""` — the fix must not turn absence of content into an error
- `/proc/sys/kernel/hostname` cross-checked against **both** `uname -n` and `.sys.info`'s `gethostname` value, so the assertion is on the right content rather than on "something was read"
- two reads of `/proc/sys/kernel/random/uuid` asserted to **differ** — that is what proves the file is really being read each time rather than a cached or truncated view of it
- a FIFO
- missing path and a directory both still raising `io`

`make test`: **3892 of 3892 passed (0 skipped, 0 failed)**. `make fuzz-smoke` and `mkdocs build --strict` clean.

Docs updated in `language/functions.md` and `reference/all-functions.md`, including the note that `.fs.size` and `read` disagreeing on a special file is expected.


singaraiona:
The blocking finding does not hold — but the first item on the non-blocking list does, and it is the more serious of the two. Both addressed below.

## Blocking finding: `.sys.exec` is not `exec_capture` here

The analysis rests on this premise:

> `.sys.exec` is `exec_capture()` in `src/ops/system.c:834-857`: `popen(cmd, "r")` followed by a `fread` loop that runs until EOF on the child's stdout

That is only true of the **two-argument** form. `ray_system_fn` branches on it (`src/ops/system.c:913`):

```c
if (!capture) {
    int rc = system(cmd);
```

`capture` is set only when a second argument `'out` is passed. The FIFO line in the test passes one argument, so it goes through `system(3)` — no pipe is created, nothing inherits a write end, and there is no EOF to wait for. The circular wait cannot form.

Confirmed directly, with the writer's reader never arriving and an external timeout:

```console
$ timeout 8 ./rayforce -i <<< '(.sys.exec "(printf ... > /tmp/fifo &) ; sleep 0")'
EXIT=0          # 124 would mean it hung
```

and the test itself, in the suite, on this branch:

```
rfl/io/read_until_eof    PASS  1.71 ms
=== 3896 of 3896 passed (0 skipped, 0 failed) ===
```

**That said, the mechanism you describe is real — it just applies to the form the test does not use.** I checked, because if it were reachable it would be a live hazard in shipped code rather than a test bug:

```console
(.sys.exec "(sleep 3 &) ; true" 'out)             -> 3.00 s
(.sys.exec "sleep 3 &" 'out)                      -> 3.00 s
(.sys.exec "(sleep 3 &) >/dev/null 2>&1" 'out)    -> 0.00 s
```

So a backgrounded child does hold the capture pipe open, exactly as argued, and `'out` will block on any command that backgrounds something without detaching its stdout. That is a genuine sharp edge in `.sys.exec 'out` (from #573, already merged) and I am filing it separately — it deserves a documented warning at minimum. Good catch, wrong call site.

## Non-blocking #1 — taken, and it was a real regression

> `(read "/dev/zero")` / `"/dev/urandom"` previously returned `""` instantly; they now consume memory until OOM.

Correct, and this is the most important finding in the report. Reproduced under a memory cap:

```
error: oom: vec_new(type=2, cap=1073741824): 1073741824 bytes
```

Turning an instant `""` into a dead process is a worse outcome than either the old behaviour or an error, and it is a regression this PR introduced.

The growth loop now stops once the next doubling would cross `ray_heap_anon_watermark()` — the heap's existing spill threshold, which resolves to physical RAM — and returns `io` naming the budget:

```
error: io: read "/dev/zero": input exceeds the readable budget of 67191541760 bytes
```

I used that watermark rather than a new constant deliberately, since a hardcoded cap is itself something this gate rejects. Stated plainly: this converts an OOM kill into a diagnosable error, it does **not** make reading an endless device cheap — it is a safety net, not a policy. If you would rather see a tighter bound, say so and I will take it, but I did not want to invent a number.

Your note that `exec_capture` has the same unbounded pattern is correct and I have left it alone here; it belongs with the `'out` backgrounding issue above.

## Non-blocking #2 and #3 — taken

The `cap * 2` overflow is covered by the same guard: `hint` is clamped to `INT64_MAX/2`, so `cap` could reach `INT64_MAX/2 + 1` and double into UB. `ncap <= 0` now fails before allocating.

The directory comment was indeed describing a now-dead fallback — the `S_ISDIR` pre-check is the real rejector. Reworded to say the in-loop guard exists for anything *else* that reports a size it cannot back.

## Non-blocking #4 — noted, not changed

`got`/`u1`/`u2` do leak on assertion failure. As you note it matches the surrounding suite, and the process exits on failure; changing it here would make this test inconsistent with its file for no behavioural gain.

`make test`: **3896 of 3896 passed**. `make fuzz-smoke` clean.

singaraiona:
Both blockers addressed in `fb900f5c`. The second one corrected my reasoning, not just my code.

## 2. The watermark is a spill threshold, not a ceiling — you are right

This is the important one, and I had it wrong. From `mem/heap.c`, as quoted:

> this never rejects work, it just picks disk over RAM

So bounding by the watermark did not bound anything — it let the loop run to physical RAM and spill the excess to disk. My previous commit message claimed it "converts an OOM kill into a diagnosable error". It did not; it traded an OOM kill for filling the spill file, which is worse. Thank you for reading the semantics rather than the name.

Now bounded by a quarter of the **remaining** headroom (`watermark - committed`), following the `ops/cdfuse.c` precedent you cited — watermark as an *input* to a derived limit, same `/4`, same "unknown RAM stays permissive" convention, still steerable through `ray_heap_set_anon_watermark`, still no new knob.

Measured:

```
error: io: read "/dev/zero": input exceeds the readable budget of 16789496832 bytes
  8.39 s   maxRSS 12496376 KB
```

Peak RSS is under the budget and below the spill threshold, so nothing reaches disk. It is bounded rather than instant, and I would rather say that plainly than overstate it again.

The budget now covers the **hinted path** too, per your non-blocking #2 — a file claiming an implausible size is refused here instead of being left to fail inside `ray_vec_new`.

The **overflow guard** was indeed written as if it caught `cap * 2` when the multiply is UB before the check runs. It now tests `cap > INT64_MAX / 2` before multiplying.

## 1. Windows — fixed, with a caveat worth recording

`mem/heap.h` is a project header and had no business inside the POSIX include guard; `struct stat` / `S_ISDIR` were used unconditionally in a function that, unlike `ray_load_file_fn`, is compiled on both paths. Both fixed.

For the record, and as your own finding notes: **no CI job builds Windows** — it is commented out at `release.yml:117`. So this was not a live break. I have fixed it because the file carries Windows arms and should stay self-consistent with them, but it is worth someone deciding whether 25 files' worth of `RAY_OS_WINDOWS` arms should be tested or removed; as it stands this gate can block on a target nothing compiles.

## Test observations — both taken, and one of them caught a bad test

> No test covers the new budget error path or the growth loop past one doubling

Correct, and worse than stated: `big.txt` has a size hint, so **every** case in that file took the single-allocation path and the `for(;;)` loop was untested.

Added a FIFO case that reads 120000 bytes — five doublings from the 4096 start — and `lang/io/read_unsized_bounded`, which drives the budget directly by moving the watermark just above what is committed, so the error path is asserted rather than assumed.

My first attempt at the FIFO case passed while testing nothing: a `format` call was missing its second argument, `mkfifo` failed, and `dd`'s `>` left an ordinary file that took the hinted path. The suite went green anyway. Verified the fix by checking the node is really a FIFO (`ls -l` type `p`) and that 120000 bytes come back.

> fixed `/tmp/rf_read_eof` path ... a collision or a failed writer subshell is an indefinite hang

Taken, and the hang risk is the sharper half of that point. Paths are now pid-suffixed via `(get (.sys.info) 'pid)`.

## Test leak — still not changed

`got`/`u1`/`u2` on assertion failure, as before: it matches the surrounding file and the process exits on failure.

`make test`: **3897 of 3897 passed**. `make fuzz-smoke` clean.

singaraiona:
Both blockers were correct and are fixed in `d9223169`.

## 2. Bounding the hinted path was a regression — and I should not have added it

You are right on both counts: a file larger than the headroom previously allocated and spilled and *worked*, and tying the limit to `anon_committed` made success depend on live process state, so the same script on the same file could pass and then fail.

Worth being explicit about how it got there, since it bears on the gate itself: **the previous audit round listed the hinted path's absence from the budget as a non-blocking inconsistency**, and I took it. This round is right and that one was not — the budget exists to terminate an *endless* stream, which is not a situation a size hint can describe. The hinted path is now unbounded again, as it was before this PR.

That also answers your point about the `cdfuse.c` comparison: you are right that the two are not analogous, because cdfuse degrades to a fallback while this produced a hard user-visible error. With the ceiling confined to the unsized path there is no fallback to want — an endless stream has no correct answer to degrade to.

## 1. The `FILE*` leak — fixed, and it was on exactly that return

Removing the hinted-path check removes the leaking early return outright. Your diagnosis of why neither test caught it is exactly right: `/dev/zero` reports size 0, so `test_read_unsized_stream_is_bounded` takes the growth-path check, which does close `fp`. The pre-check was the only unclosed exit in the function and the only one nothing exercised.

## Non-blocking — all four taken

**Peak is `3*cap`, not `2*cap`.** Correct — the old buffer is alive across the `memcpy`. The guard now bounds the peak (`cap * 3 > budget`) rather than the result.

**`budget of 0 bytes`.** Split into two messages: `input too large to read` for the overflow case, `unsized input exceeds the readable budget of N bytes` for the budget case. Overflow is now checked against `INT64_MAX/3` before the multiply.

**`(long)(INT64_MAX/2)` on ILP32.** Now compares in `int64_t`.

**Watermark restore pinning the sentinel.** Good catch — the test restored the *effective* value, so where the suite had `0` ("use physical RAM") it would have left the RAM figure behind. It now detects that case and restores the sentinel.

**FIFO writer scheduling.** Acknowledged and not fixed: the `.rfl` harness has no timeout and I would rather not invent one here. Pid paths address collisions, not a writer that never starts. If you would like it gone entirely, the growth path is also covered by `lang/io/read_unsized_bounded` in C, so the FIFO cases could be dropped without losing coverage of the loop — say the word.

## New test

`lang/io/read_unsized_bounded` now also asserts that a 2 MB **sized** file reads fine under a ~1 MB budget, which pins precisely the regression you identified. Mutation-verified — re-adding the hinted-path check fails it:

```
lang/io/read_unsized_bounded    FAIL
=== 3896 of 3897 passed (0 skipped, 1 failed) ===
```

Docs now state that an unsized stream can be refused by size and that a sized read cannot, which was the omission you flagged.

`make test`: **3897 of 3897 passed**. `make fuzz-smoke` and `mkdocs build --strict` clean.

PR #583, opened 2026-09-19
fix(if): translate only the FILE-domain cells an if branch reads

ser-vasilich:
## Bug

An `if` whose branch scans a SYM column with a FILE-domain vocabulary translates that branch's cells into runtime ids through the domain's whole-vocabulary LUT (`ray_sym_domain_runtime_lut`). The first request interns every vocabulary entry into the global symbol table, with the dotted-name split, and keeps it for the life of the process.

On a 100M-row table with a 19.7M-entry column the branch read 1.7M distinct values and the translation cost about 45 GB of permanent symbol-table growth. Interning the vocabulary also dominated the query's run time.

## Fix

The branch carries its own translation table (`if_sym_side_t`, `src/ops/pivot.c`): slots start unresolved and a position is interned the first time the branch reads it, from the raw vocabulary snapshot (`ray_sym_domain_raw_pin`). The ids produced are exactly the ones the LUT would have returned, so results are unchanged.

- The serial scatter path (`if_scatter_sym`) resolves on the fly.
- The parallel eager fill pre-resolves the positions it will read (`if_sym_side_prepare`, serial) before dispatching, so workers only read the table — the frozen-table rule of `sym.c` still holds.
- Runtime-domain, STR and scalar branches need no table and are untouched.

Memory for the table is one `int64` per vocabulary position for the duration of the `if`; symbol-table growth is bounded by the values the branch actually reads.

## Test

`test/rfl/symbol/file_domain_if.rfl`: `if` over a splayed (FILE-domain) SYM column against the same expressions over the in-memory table — trivial branches through the pooled fill (70000 rows, past the parallel threshold), a computed branch through the selected path, the column on both sides under a row selection, as a group key, and the empty symbol on either side.


singaraiona:
Reviewed and merging. The shape is right: interning is confined to the calling thread, the parallel path pre-resolves through `if_sym_side_prepare` before dispatch so workers only read, and that is what keeps `sym.c`'s frozen-table rule intact. Bounding symbol-table growth by what the branch actually reads, rather than by the vocabulary size, is the correct fix — 45 GB of permanent growth for 1.7M values read is a good catch.

One non-blocking note for a follow-up, on the failure mode rather than the logic.

`if_sym_side_lookup` maps an unprepared position to `0` — the empty symbol — and the comment is explicit that this stands in for "a bug in the caller's prepare pass". That trades a crash for silently wrong data: a prepare pass that missed positions produces a result full of empty symbols that looks well-formed and would pass any test not asserting on those exact rows.

Since the prepare/lookup split is the whole safety argument for the parallel path, it is worth making a prepare miss loud in DEBUG — an assert there costs nothing in release and would turn a future refactor's mistake into a failing test instead of a wrong answer. The release-build fallback can stay exactly as it is.

Not blocking the merge.

PR #584, opened 2026-09-19
perf(group): evaluate a derived symbol key in chunks over the vocabulary bytes

ser-vasilich:
## Bug

When a group key is computed from one SYM column (`derived_key_over_sym_domain`), the expression is evaluated once per distinct symbol against a mini-table holding the distinct ids. For a column whose ids live in a FILE domain every string step of that expression became a permanent cost:

- `substr` / `str-find` intern each distinct intermediate result into the global runtime symbol table (with the dotted-name split, about 1 KB per URL-shaped value);
- an `if` that keeps the column itself translates its cells into runtime ids as well;
- resolving the input cells materialises an atom per vocabulary entry.

On a 100M-row table with a 19.7M-entry column a five-step key expression left about 65 GB in the symbol table that never came back, and the process peaked above 70 GB.

## Fix

`derived_key_str_chunks` (`src/ops/query.c`): when the expression's result over the SYM column is SYM and the column's vocabulary is readable through the raw snapshot, the referenced column is presented to the same DAG as a STR column built from the vocabulary bytes, two million distinct values at a time. Every intermediate is then a chunk-sized STR vector that dies with its chunk, and only the final key strings are interned — once per distinct value, giving the ids the SYM evaluation would have produced. The result type is checked by compiling the expression against the SYM column first, so the key column keeps its type; any other shape (non-SYM result, runtime-domain column, no raw snapshot) keeps the one-shot SYM evaluation.

The same query now peaks below 8 GB with identical output. `RAY_DERIVED_KEY_CHUNK` overrides the chunk length so a test can drive several chunks through a small vocabulary.

## Test

`test/rfl/group/derived_key_sym_domain.rfl`, new section: the existing key expressions over a splayed (FILE-domain) copy of the table against the in-memory oracle, with the default chunk and with a 100-value chunk (several chunks over the vocabulary), a non-symbol key result (`strlen`) that keeps the one-shot path, and the null-bearing column.


singaraiona:
The fix itself is right and the win is large — 65 GB of permanent symbol-table growth down to a peak under 8 GB, with identical output, by keeping intermediates as chunk-scoped STR vectors and interning only the final key strings. No objection to any of that.

One thing to change before this goes in: **`RAY_DERIVED_KEY_CHUNK` is a new env-var knob**, and we removed the last one of those on purpose.

```c
static int64_t derived_key_chunk_rows(void) {
    const char* env = getenv("RAY_DERIVED_KEY_CHUNK");
```

`RAY_DIRECT_CACHE_MB` was the same shape — a tunable added for internal reasons — and it came out in `1ae57b98`. The standing rule is that configuration surface stays small and discoverable: internal mechanisms get derived defaults, and where a test needs to steer one, it goes through a C API rather than the environment. `ray_heap_direct_cache_drain` is the precedent, and `ray_ipc_set_auto_journal_eval_for_test` (`core/ipc.h:98`, `#ifdef DEBUG`) is the same idea for a test-only seam.

Reading the PR description, the override exists *only* so a test can drive several chunks through a small vocabulary — there is no operator use case. So a `#ifdef DEBUG` setter gives you exactly what the test needs, keeps it out of release builds entirely, and avoids a `getenv` on every call to `derived_key_chunk_rows()`.

Two smaller points while you are in there:

`DERIVED_KEY_CHUNK` is `1 << 21` with no derivation. If there is a reason it is two million rather than any other power of two, the comment should say what it is; otherwise it would be better tied to something that already exists — `RAY_MORSEL_ELEMS` / `RAY_DISPATCH_MORSELS` in `ops/ops.h` are the granularity budgets the rest of the engine sizes against.

The test sets and then clears the variable with `.os.setenv`, which leaves process-global state behind if an assertion between those two lines fails. A DEBUG setter with the same save/restore discipline avoids that too.

Everything else — the type probe compiling against the SYM column first so the key column keeps its type, the fallbacks for non-SYM results and runtime-domain columns, and `73265b50` dropping the extra retain that was pinning the domain — looks correct to me. Happy to merge as soon as the knob is gone.

